PR #27, opened 2026-01-09
Export tweaks

krisajenkins:
Following the vim-style keybindings, `y` now yanks to the clipboard in the same way `:clip` does. And if there's nothing to go to the clipboard, you get a warning message.


agavra (approved):
great idea

Issue #2, opened 2026-01-08
Support modifying comments


agavra:
Right now it's all or nothing - delete a comment with `dd` and create a new one is your only option. We should support something like `m` or `i` to edit a comment.

PR #37, opened 2026-01-09
Support modifying comments

YPares:
This adds an 'e' command to edit a comment, if there is one next to the cursor. Key added to help menu.

Closes https://github.com/agavra/tuicr/issues/2

### DIFF
--- a/src/ui/help_popup.rs
+++ b/src/ui/help_popup.rs
@@ -106,7 +106,7 @@ pub fn render_help(frame: &mut Frame) {
         ]),
         Line::from(vec![
             Span::styled(
-                "  e         ",
+                "  i         ",
                 Style::default().add_modifier(Modifier::BOLD),
             ),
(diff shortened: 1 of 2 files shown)


src/ui/help_popup.rs
```diff
@@ -104,6 +104,20 @@ pub fn render_help(frame: &mut Frame) {
             ),
             Span::raw("Add file comment"),
         ]),
+        Line::from(vec![
+            Span::styled(
+                "  e         ",
```
agavra: I think we should consider using `i` or `a` to match vim "enter insert mode". `e` is typically reserved for moving forward when selecting words

agavra (commented):
Nice thanks for the PR @YPares! I'll test this locally in a little bit - just one comment about the actual binding to use

agavra (approved):
Tested it locally and it works great! Just waiting to change the binding and I'll merge

PR #47, opened 2026-01-11
feat: add scrolling support for file list panel

airscholar:
## Summary
- Fixes file list panel not scrolling when there are many uncommitted files
- Files beyond the visible viewport were previously inaccessible

## Problem
When reviewing a large number of uncommitted files (e.g., 60+), the left file list panel did not scroll. Files that exceeded the viewport height were simply cut off with no way to view or navigate to them.

## Solution
Replaced the static `Paragraph` widget with ratatui's `List` widget and `ListState`, which provides automatic viewport scrolling. The list now scrolls to keep the selected file visible when navigating with `j`/`k` keys.

## Changes
- **`src/app.rs`**: Updated `FileListState` to wrap ratatui's `ListState` with helper methods for selection
- **`src/ui/app_layout.rs`**: Switched from `Paragraph` to `List` widget with `render_stateful_widget`
- **`src/main.rs`**: Updated to use new `selected()` method

## Testing
1. Create or checkout a branch with many uncommitted files (50+)
2. Run `tuicr`
3. Press `Tab` to focus the file list panel
4. Navigate with `j`/`k` - the list should scroll to keep the selection visible

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -85,16 +85,9 @@ pub struct App {
     pub supports_keyboard_enhancement: bool,
 }
 
+#[derive(Default)]
 pub struct FileListState {
     pub list_state: ratatui::widgets::ListState,
-}
-
-impl Default for FileListState {
-    fn default() -> Self {
-        Self {
-            list_state: ratatui::widgets::ListState::default(),
-        }
-    }
 }
 
 impl FileListState {


agavra:
Tested this locally and confirmed it works. I think maybe a better solution is to automatically reject the patch and tell Claude to break it up into smaller commits 🤣 (just kidding).

Thanks for the contribution @airscholar and for catching this bug!

PR #50, opened 2026-01-11
feat: Add hierarchical file tree with expand/collapse

airscholar:
## Summary
- Add hierarchical file tree view with expand/collapse functionality
- Use Space to select files and toggle directory expansion
- Use `o`/`O` to expand/collapse all directories
- Underline styling for selected items in file list
- Tree structure with ▾/▸ icons for expanded/collapsed directories
- Auto-expand parent directories when jumping to file
- Sync selection between file list and diff panels
- Add 7 unit tests for tree functionality

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::path::PathBuf;
 
 use crate::error::{Result, TuicrError};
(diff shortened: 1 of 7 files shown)


agavra (commented):
I love this! So much so I think we should make it the only option. Some feedback inline

PR #54, opened 2026-01-11
fix: support wayland clipboard. update arboard dependency to include wayland-data-control feature

martintrojer:


Cargo.toml
```diff
@@ -33,7 +33,7 @@ anyhow = "1.0"
 directories = "5.0"
 unicode-width = "0.2"
 uuid = { version = "1.0", features = ["v4"] }
-arboard = "3.4"
+arboard = { version = "3.4", features = ["wayland-data-control"] }
```
agavra: I was just going to comment that I prefer keeping it simple and just enabling it globally. I checked and it doesn't add any meaningful amount to the binary size. :)

agavra (approved):
Thanks for the patch! I just did a patch release this morning with some bug fixes, do you need this one released? Otherwise I'll wait to batch together a few more fixes before doing another release.

martintrojer:
> Thanks for the patch! I just did a patch release this morning with some bug fixes, do you need this one released? Otherwise I'll wait to batch together a few more fixes before doing another release.

No rush, using source build locally for now.

> I was just going to comment that I prefer keeping it simple and just enabling it globally. I checked and it doesn't add any meaningful amount to the binary size. :)

I had  Claude double check this and apparently there is a platform gate in the library so this should go away completely on non-linux builds.

Issue #48, opened 2026-01-11
consider always showing the diff selector menu


agavra:
right now the behavior is to only show the git hash diff selector if there are no uncommitted changes. after we implement #43 we should consider always showing the diff selector so unstaged changes can be viewed alongside old(er) commits

agavra:
Fixed in #159

Issue #43, opened 2026-01-10
multi-commit view doesn't work


agavra:
when you select multiple commits in the multi-select view at startup it doesn't let you press enter

PR #66, opened 2026-01-12
Multi-Session and Multi-Repository Support with Performance Optimizations

### LINKED ISSUE #5: handle multiple sessions / multiple repositories
This is a big one. Currently if the base commit is different from whatever review you were last working on it will just nuke the history for your old review. That sucks! We should make sure that we have some mechanism for mapping reviews to the _current_ context.

avifenesh:
# Multi-Session and Multi-Repository Support

fixes #5

## Overview

This PR updates session persistence to support multiple review contexts (branch + diff source) without deleting existing sessions. It also fixes commit-range session persistence so reopening the same commit selection restores comments, and hardens session filename parsing so branch names with underscores or hex-like segments do not prevent loading.

## Key Changes

- Review sessions now record branch, diff source, and commit range (schema v1.2) with backward-compatible defaults.
- Session loading is context-aware (repo fingerprint + diff source + branch + commit range) and updates base commit non-destructively.
- Session filenames include repo fingerprint/branch/diff source; parsing tolerates underscores and hex-like branch segments.
- Added commit-range persistence tests and filename parsing regression coverage.

## Tests

- `cargo test`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,8 +34,6 @@ use handler::{
     handle_search_action, handle_visual_action,
 };
 use input::{Action, map_key_to_action};
-use output::markdown::export_to_clipboard;
-use persistence::storage::save_session;
 use theme::{parse_cli_args, resolve_theme};
 
 /// Timeout for the "press Ctrl+C again to exit" feature
--- a/src/output/markdown.rs
+++ b/src/output/markdown.rs
@@ -301,7 +301,12 @@ mod tests {
     #[test]
     fn should_fail_generate_export_content_when_no_comments() {
         // given
-        let session = ReviewSession::new(PathBuf::from("/tmp/test-repo"), "abc1234def".to_string());
+        let session = ReviewSession::new(
+            PathBuf::from("/tmp/test-repo"),
+            "abc1234def".to_string(),
+            Some("main".to_string()),
+            SessionDiffSource::WorkingTree,
+        );
         let diff_source = DiffSource::WorkingTree;
 
         // when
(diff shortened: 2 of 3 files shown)


src/input/keybindings.rs
```diff
@@ -124,6 +125,7 @@ fn map_normal_mode(key: KeyEvent) -> Action {
         // Mode changes (use _ for shifted characters like : and ?)
         (KeyCode::Char(':'), _) => Action::EnterCommandMode,
         (KeyCode::Char('?'), _) => Action::ToggleHelp,
+        (KeyCode::Char('m'), KeyModifiers::NONE) => Action::EnterCommitSelectMode,
```
agavra: let's punt this to a separate PR since this is already a big PR, and also consider #48 instead

src/model/review.rs
```diff
@@ -39,26 +39,50 @@ impl FileReview {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SessionDiffSource {
+    WorkingTree,
+    CommitRange,
+}
```
agavra: we should consider how this will work if we allow selecting commits + working tree (e.g. I want to see the diff from the last commit AND the current working tree). also see #43 

we shouldn't fix this in this PR, but the design should allow for it

agavra (commented):
This is _awesome_ @avifenesh thanks for the detailed design and contribution! 🚀 🔥 🤩 

I left some comments inline, but for the most part I'd like to split this PR into multiple so that we're not jamming a bunch of unrelated changes into one thing.

agavra:
@avifenesh I would like to continue with this one, I haven't had time to give it another round of review after you split out the other stuff. Because it's a pretty substantial PR I'll need to do a heavy round of testing and review. Did you update this PR after splitting out the other stuff?

avifenesh:
Will do, wanted to know that it is still relevant 👍🏽 @agavra

PR #74, opened 2026-01-13
feat: add Ctrl+P support for pasting from clipboard

Sarthakkeche:
Description
Summary This PR adds support for pasting text from the system clipboard into the comment input box using Ctrl+P. Previously, users had to manually type out long comments or references, which was inefficient.

Changes

Clipboard Support: Added the arboard crate to enable cross-platform system clipboard access.

Event Handling: Implemented a new event listener in main.rs to detect Ctrl+P when in InputMode::Comment.

Windows Fix: Added a check for KeyEventKind::Press to filter out key release events, resolving a double-input bug common on Windows terminals.

Sanitization: Added logic to strip newlines (\r\n and \n) from pasted text to prevent accidental command execution or UI breakage.

Testing

Environment: Tested on Windows 11 (PowerShell/VS Code Terminal).

Scenario: Copied text from an external browser and successfully pasted it into the comment box using Ctrl+P.

Verification: Confirmed that text appears once (no double typing) and appends correctly to the existing buffer.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,16 +12,12 @@ mod ui;
 use std::io;
 use std::time::Duration;
 
-use arboard::Clipboard; 
+use arboard::Clipboard;
 
 use crossterm::{
-    event::{
-        self, Event, KeyCode, KeyModifiers, KeyEventKind,
-    },
+    event::{self, Event, KeyCode, KeyEventKind, KeyModifiers},
     execute,
-    terminal::{
-        EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode,
-    },
+    terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
 };
 use ratatui::{Terminal, backend::CrosstermBackend};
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -63,108 +56,109 @@ fn main() -> anyhow::Result<()> {
...
-            // Handle pending d command for dd delete comment
-            if pending_d {
-                pending_d = false;
-                if key.code == crossterm::event::KeyCode::Char('d') {
-                    if !app.delete_comment_at_cursor() {
-                        app.set_message("No comment at cursor");
+                // --- PASTE LOGIC (CTRL+P) ---
+                if app.input_mode == InputMode::Comment
+                    && key.code == KeyCode::Char('p')
+                    && key.modifiers.contains(KeyModifiers::CONTROL)
```
agavra: on mac the correct key binding is `command+p`, I wonder if there's a way for us to improve the integration with the native keyboard instead of hard-coding `ctrl-p`

this might be a little trickier than I had originally hoped, let me know what you find

agavra (commented):
thanks @Sarthakkeche - I tested it locally and it indeed works with `ctrl+p` but the problem is that copy-paste commands are platform dependent, we need a way to figure out how to handle that appropriately

PR #75, opened 2026-01-13
fix: enforce scroll bounds to prevent scrolling past content

airscholar:
## Summary
- Fix horizontal scrolling allowing users to scroll infinitely right past visible content
- Fix vertical scrolling allowing users to scroll past the bottom of the diff
- Add proper bounds checking based on actual visible content dimensions

## Changes
- Added `viewport_width` and `max_content_width` fields to `DiffState` and `FileListState`
- Calculate max content width from **visible lines only** during render
- Auto-clamp `scroll_x` when content changes or after vertical scrolling
- Apply consistent `viewport_height.max(1)` handling to prevent edge cases

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -154,6 +154,8 @@ pub struct App {
 pub struct FileListState {
     pub list_state: ratatui::widgets::ListState,
     pub scroll_x: usize,
+    pub viewport_width: usize,    // Set during render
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thanks @airscholar this looks good to me!

PR #82, opened 2026-01-14
#15 Fix double keystroke bug in Windows

### LINKED ISSUE #15: verify this works on windows / linux + not iTerm2
I've only developed this on Mac OS + iTerm2 so we should verify that the TUI works on other terminals and operating systems

imbatcat:
Added a check for KeyEventKind::Press to ensure that only key press events are processed, since Windows also sends KeyEventKind::Release, which will results in another keystroke. 

fixes #15

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,7 +14,7 @@ use std::time::Duration;
 
 use crossterm::{
     event::{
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thanks for the fix @imbatcat 🙏

PR #91, opened 2026-01-15
feat: add manual commit selection mode

avifenesh:
## Summary
Add the ability to manually enter commit selection mode using the `m` keybinding, allowing users to review commits even when there are working tree changes present.

## Changes
- Added `EnterCommitSelectMode` action and `m` keybinding in normal mode  
- Added `enter_commit_select_mode()` and `exit_commit_select_mode()` methods
- Changed Esc behavior in commit select mode to exit back to normal mode (returns to working tree if applicable)
- Updated status bar hints: `m:commits` in normal mode, `Esc:back` in commit select mode
- Added "Review Modes" section to help popup documenting the feature

## Usage
1. Press `m` in normal mode to enter commit selection mode
2. Browse recent commits (20 most recent) using `j`/`k`
3. Select commits with `Space`, confirm with `Enter`
4. Press `Esc` to exit back to normal mode

## Context
This PR implements the deferred feature from PR #66 review feedback, allowing manual entry into commit selection mode.

Related: #48

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1699,7 +1699,7 @@ impl App {
         self.session =
             ReviewSession::new(self.vcs_info.root_path.clone(), newest_commit_id.clone());
 
-        // Add files to session (only if they don't already exist)
+        // Add files to session
         for file in &diff_files {
             let path = file.display_path().clone();
             self.session.add_file(path, file.status);


avifenesh:
> If I'm being totally honest, I'm not 100% sure this feature is helpful or if it's just a little distracting. If we implement #48 and allow you to select both the working tree and various commits then why would you want to go back to commit mode instead of starting a new session?
> 
> Thanks for understanding the back and forth, just want to make sure we're building the right thing here.

TBH it exists because it was in the previous PR and moved to here. I generally wanted to support multi-sessions. :) 
This needs more thoughts, I guess.

Based on my CC usage: 
Always in the worktree(unless I'm working only one session, which is rare, and even then probably worktree, exceptions are hot fixes or small things like README updates), I usually care about the diff from origin main, origin, or diff from another checkpoint. It won't be chosen commits; it should be, for me, choosing which diff to show compared to what point.
I would want the default behavior to be to open a diff from the origin main. Typically what I care about, ill navigate inside it. Some time, I asked for a fix, so I want the last commit. But I probably want it as part of the prev session, so I want to open the session I asked for changes in and then change “view” to “show me the difference compared to this commit”.  But I want it to be a continuous session of review. 
@agavra

agavra:
got it, that makes sense to me thanks for clarifying your workflow! let's modify the command to be `:commits` and then it'll LGTM

PR #96, opened 2026-01-15
feat: add optional Jujutsu (jj) support [3 of 4]

martintrojer:
Add jj backend using CLI commands, similar to hg support. Unify hg and jj diff parsers into shared diff_parser module with DiffFormat enum to handle format-specific differences.

Detection order: jj → git → hg (jj first since it's git-backed).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - run: cargo check
-      - run: cargo check --features hg
+      - run: cargo check --features hg,jj
 
   fmt:
     name: Format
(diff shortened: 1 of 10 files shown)


agavra:
I'll merge this after we fix `clippy` problems

martintrojer:
Wait, is the error specifically on an empty revision ? Maybe there is a problem in that case

YPares:
@martintrojer 

> Wait, is the error specifically on an empty revision ? Maybe there is a problem in that case

Yes, I get it specifically when I'm on an empty revision.

YPares:
Some visual aspects could change if using a JJ repo instead of a "bare" Git repo:
- showing change ids along commit ids (or instead of commit ids)
- showing branch names even if on detached HEAD (I saw this was already done), as JJ has no notion of "current branch" (branches are just aliases for revisions which don't move by themselves, basically they are just there for interop with Git remotes)

Issue #7, opened 2026-01-08
support light/dark mode


agavra:
light mode colors are horrible right now. if anyone uses light mode they might have a mini-heart attack.

YPares:
Another option could be to stick to ANSI 16 color palette only (for starters), BUT it would require some special care re. red/green backgrounds (it would be more limiting).

`bat` is pretty legible with that (only ever used it this way, with `--theme=ansi`), again modulo the caveat that it doesn't have to deal with red/green background because it's not showing diffs. But it would require no special handling from `tuicr` or the user (who already has a terminal with a color palette set).

Issue #44, opened 2026-01-10
Consider custom color themes


agavra:
Color is a very personal thing, we should allow people to customize their color themes instead of depending on the ones that ship with tuicr

PR #105, opened 2026-01-16
ui: add theme support with dark and light modes

Mic92:
Users with light terminal backgrounds cannot read the current color scheme. The hardcoded dark theme colors have poor contrast on light backgrounds.

Add a --theme flag to select between dark (default) and light themes:

  tuicr --theme dark   # default, original colors
  tuicr --theme light  # optimized for light terminal backgrounds

Introduce a Theme struct that holds all color definitions and thread it through the rendering code. All style functions in src/ui/styles.rs now take a &Theme parameter instead of using hardcoded constants.

The light theme uses subtle backgrounds (very light green/pink) with dark text for good contrast on light terminal backgrounds.

Closes: #7

### DIFF
--- a/src/theme/mod.rs
+++ b/src/theme/mod.rs
@@ -2,11 +2,17 @@
 //!
 //! Provides dark and light themes with automatic terminal background detection.
 
+use std::sync::OnceLock;
+
 use ratatui::style::Color;
 
+use crate::syntax::SyntaxHighlighter;
+
 /// Complete color theme for the application
-#[derive(Debug, Clone)]
 pub struct Theme {
+    /// Cached syntax highlighter (lazily initialized)
+    highlighter: OnceLock<SyntaxHighlighter>,
+
     // Base colors
     pub bg_highlight: Color,
     pub fg_primary: Color,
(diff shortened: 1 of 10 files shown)


src/theme/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+/// Parse --theme argument from command line
+pub fn parse_theme_arg() -> ThemeArg {
+    let args: Vec<String> = std::env::args().collect();
+
+    for i in 0..args.len() {
+        // Handle --theme value
+        if args[i] == "--theme" {
```
Mic92: clap always feels like overkill (all the crates it adds, while not saving that much code to be honest), but let me know if you prefer this over a handrolled cli parser. In case you are ok, with the hand-rolled cli, I can also add --help.

agavra:
Really happy to see this change come in! I don't think this fixes #44 quite yet since users can't define their own without modifying source, but it opens the door to that.

I think we need to also change https://github.com/agavra/tuicr/blob/10802b89f447d86a6f351a0321a29fedb764447b/src/syntax/mod.rs#L110-L125  theme for light otherwise we get:
image

I also think we should change the syntect theme as well to respect this as well (e.g. use `base16-ocean.light` instead of `"base16-eighties.dark"`)

Would be good to share screenshots of your final preview with it before we merge.

PR #106, opened 2026-01-16
fix: Replace tabs with space

manishgjain:
When reviewing diffs in languages which use tabs instead of space like
golang, tabs are not rendered which makes it hard to review the diff.
This change attempts to fix this by replacing tabs with 4 spaces.
Possible future extension could be to let users customize the tab
width via configuration.

Before
Screenshot 2026-01-16 at 13 40 33

After
Screenshot 2026-01-16 at 13 40 59


agavra (approved):
Thanks for the contribution @manishgjain - this LGTM

PR #117, opened 2026-01-18
Mouse Improvements

### LINKED ISSUE #98: Mouse scrolling should move the viewport, not the cursor
`tuicr` supports mouse input, and that's great, though it has the strange fzf-y behaviour that scrolling moves the cursor, not the viewport.
Is it something that could be changed, or is it some behaviour embedded in ratatui that `tuicr` doesn't have a say about?

YPares:
Closes #98 

This implements a more natural mouse scrolling management, where the viewport scrolls instead of the cursor.
Also, besides Tab, panels can be selected via mouse clicking.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,8 +15,8 @@ use std::time::{Duration, Instant};
 
 use crossterm::{
     event::{
-        self, Event, KeyEventKind, KeyboardEnhancementFlags, PopKeyboardEnhancementFlags,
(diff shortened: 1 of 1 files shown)


agavra (approved):
very nice, curious if this works in ssh mode but that's not a blocker for merging it.

by the way @YPares your engagement on this project has been great, would you want to be a committer? I can give you write access and add to the CONTRIBUTING.md a committer section. I don't _require_ PR approval but for anything a little controversial it would be nice to get a +1.

PR #120, opened 2026-01-19
fix unconditional exit with ":commits" command

agavra:


avifenesh:
i didnt finish testing haha
i guess you validated base on the follow up pr? @agavra

PR #122, opened 2026-01-20
feat: add vim-style warning on exit with unsaved changes

### LINKED ISSUE #61: issue a warning before exiting from :q if there are comments in the buffer
tuicr should issue a warning before exiting from :q if there are comments in the buffer to protect against accidentally exiting after reviewing.

mikecfisher:
## Summary
- `:q` shows error "No write since last change (add \! to override)" when there are unsaved changes
- `:q!` forces quit regardless of dirty state  
- Pressing `q` in normal mode warns once ("Unsaved changes. Press q again to quit."), second `q` quits
- Warning resets when any other action is taken

Closes #61

## Test plan
- [ ] Add a comment, press `q` → should show warning
- [ ] Press `q` again → should quit
- [ ] Add a comment, type `:q` → should show error message
- [ ] Type `:q!` → should quit immediately
- [ ] Add a comment, save with `:w`, press `q` → should quit without warning

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -150,6 +150,7 @@ pub struct App {
 
     pub should_quit: bool,
     pub dirty: bool,
+    pub quit_warned: bool,
     pub message: Option<Message>,
     pub pending_confirm: Option<ConfirmAction>,
     pub supports_keyboard_enhancement: bool,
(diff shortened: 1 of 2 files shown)


agavra (approved):
Looks great for your first rust contribution @mikecfisher - thanks for the fix! Also tested locally to make sure it works.

PR #124, opened 2026-01-20
fix: hg/jj parse paths from rename/copy metadata and binary file lines

martintrojer:
fix: parse paths from rename/copy metadata and binary file lines
This fixes a panic "DiffFile must have at least one path" when reviewing
commits with:
- Pure renames/copies without content changes (no ---/+++ lines)
- Binary files in git-style diff format

Changes:
- Extract paths from "rename from/to" and "copy from/to" lines in parse_file_header()
- Add parse_binary_file_line() to handle git format ("Binary files a/... and b/... differ")
  and hg format ("Binary file ... has changed")
- Rename git_should_* tests to jj_should_* since this code is for jj/hg, not git
- Add integration tests for jj: renamed files, binary added/deleted
- Add integration tests for hg: renamed files, copied files, binary added/deleted

### DIFF
--- a/src/vcs/diff_parser.rs
+++ b/src/vcs/diff_parser.rs
@@ -127,13 +127,33 @@ where
         } else if line.starts_with("deleted file") {
             status = FileStatus::Deleted;
             lines.next();
-        } else if line.starts_with("rename from") {
+        } else if let Some(path) = line.strip_prefix("rename from ") {
             status = FileStatus::Renamed;
+            old_path = Some(PathBuf::from(path));
             lines.next();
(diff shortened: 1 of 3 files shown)


agavra (approved):
@martintrojer these is awesome test coverage, I was always a little worried about rolling our own diff parser but this gives me more confidence in it. thank you!

Also you've had some consistent high quality contributions, if you'd like I'm happy to include you as a committer to `tuicr` 🥳 there's no official responsibility, you'd just be given write access to the main repository and your name would be on `CONTRIBUTING.md` as one of the committers.

PR #154, opened 2026-02-06
Add syntax highlighting for diffs

mikecfisher:
This adds proper syntax highlighting to diff views. There was already some scaffolding for it in the codebase, but it didn't actually work — the highlighter was being fed every diff line (context, additions, deletions all mixed together) in a single pass, so syntect's parser state would fall out of sync whenever additions and deletions were interleaved in a hunk. In practice that meant most real-world diffs rendered without any colors.

I rewrote the highlighting path to split the diff into two clean sequences before running syntect: one for the old file side (context + deletions) and one for the new file side (context + additions). Each side gets its own highlighting pass against coherent source, then the spans get mapped back to the original diff line order. Added and deleted lines also get their respective background colors applied on top of the syntax colors.

I also expanded the syntax detection quite a bit so more file types get highlighted:

- Case-insensitive extension matching (e.g. `MAIN.RS`)
- Shebang detection for extensionless scripts
- Fallback mappings for extensions syntect doesn't ship with — TypeScript, Vue, Svelte, SCSS, Dockerfile, Prisma, Elixir, Nim, Kotlin, and a bunch more

All existing tests pass plus I added a bunch of new ones covering the split logic, per-line failure isolation, the fallback mappings, and the background color application.

## Test plan
- [x] `cargo test` — 174 tests pass
- [x] Manually reviewed interleaved TypeScript hunks that were previously broken

### DIFF
--- a/src/syntax/mod.rs
+++ b/src/syntax/mod.rs
@@ -5,6 +5,12 @@ use syntect::parsing::SyntaxSet;
 
 use crate::model::diff_types::LineOrigin;
 
+/// A single line of highlighted spans (style + text pairs).
+type HighlightedSpans = Vec<(Style, String)>;
+
+/// Per-line highlight results for a file: `Some` if the line was highlighted, `None` on failure.
+type HighlightedLines = Vec<Option<HighlightedSpans>>;
+
 /// Helper to highlight lines of code from a diff
 pub struct SyntaxHighlighter {
     pub syntax_set: SyntaxSet,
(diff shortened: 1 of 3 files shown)


agavra:
Nice! Thanks @mikecfisher - I'll review this in a bit the approach makes a lot of sense to me much better than what's there.

Until then, let's split out the `vendored-libgit2` to another change and feature guard it so that you have to install from cargo if you want to use that.  IIRC, vendoring libgit2 depends on openssl which causes a bunch of problems when building binaries for various systems.

mikecfisher:
Thanks @agavra . Good call makes total sense. PR updated!

PR #155, opened 2026-02-07
Replace syntect defaults with two-face for expanded syntax highlighting

mikecfisher:
Follow-up to #154.

That PR did a few things: it fixed the diff highlighting to split by old/new side (the old code fed interleaved lines through syntect's stateful parser), added per-line failure isolation so one bad line doesn't kill highlighting for the whole file, and included a block of fallback extension mappings to cover languages syntect doesn't ship with (ts -> js, toml -> yaml, kt -> java, etc.).

The first two changes are good and still in the codebase. But while working on the fallback mappings I realized the root problem wasn't missing mappings, it's that syntect's built-in syntax set only covers 40 languages. Mapping TypeScript to JavaScript isn't proper syntax highlighting

The [`two-face`](https://github.com/CosmicHorrorDev/two-face) crate bundles the syntax and theme definitions from `bat`, which gives us ~170 languages natively — TypeScript, TSX, TOML, Dockerfile, Kotlin, Swift, Zig, Svelte, Vue, Elixir, and everything else that was in the fallback table. It returns the same syntect types, so the rest of the highlighting pipeline is untouched.

What changed:
- `SyntaxSet::load_defaults_newlines()` -> `two_face::syntax::extra_newlines()`
- `ThemeSet::load_defaults()` -> `two_face::theme::extra()`
- `SyntaxHighlighter::new()` takes `EmbeddedThemeName` instead of `&str` (compile-time checked)
- Removed the fallback extension mapping block since two-face covers all of those natively
- Removed `syntect-tui`, which was never imported anywhere and was pulling in an old `ratatui 0.28.1`

The `syntect` crate itself stays — two-face wraps it, it doesn't replace it.

## Test plan
- [x] `cargo build` compiles clean
- [x] `cargo test` — all 159 tests pass
- [ ] Open a TypeScript/TSX diff and confirm native highlighting (not JS fallback)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,7 +95,16 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
 dependencies = [
- "bit-vec",
+ "bit-vec 0.6.3",
+]
+
+[[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec 0.8.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thanks @mikecfisher :)

PR #162, opened 2026-02-10
Add config file support and Catppuccin themes

### LINKED ISSUE #131: Add Theme Support and Syntax Highlighting
Hey @agavra I added catpuccin themes and syntax highlighting to my local fork. If you're interested in this I'm happy to open a PR :) 

Image

mikecfisher:
Closes #131

Hey! Here's the PR we talked about in that issue. I went with TOML for the config since you mentioned you preferred that over JSON. The config file lives in the standard XDG path (`~/.config/tuicr/config.toml`) on Linux/macOS and `%APPDATA%\tuicr\config.toml` on Windows, like we discussed.

Right now the only config option is `theme`, but the structure is there for adding more stuff down the road. I had to redo this from scratch since my earlier approach didn't really fit with the current codebase, but I think it came out cleaner this time around.

Here's what's in here:

- New `src/config/mod.rs` module for finding and loading the TOML config
- Four Catppuccin themes: latte, frappe, macchiato, mocha
- `--theme` flag now accepts the new theme names alongside the existing dark/light ones
- Theme resolution: CLI flag beats config file, config file beats the default (dark)
- Added new color fields to the Theme struct so message badges, status indicators, help text, and panel backgrounds are all theme-aware now
- All panels and popups use the theme's panel background color instead of relying on terminal defaults, so everything looks consistent
- Tests for config loading and theme resolution
- Updated the README with config docs

Only new dependency is the `toml` crate.

Invalid theme names on the CLI cause an immediate exit with an error, and unknown keys in config.toml are rejected too, so nothing fails silently. I tested across iTerm2, Kitty, and Alacritty and it looks good on all of them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,8 +13,12 @@ tuicr is a Rust terminal UI application for code reviews. Here's the module stru
 ```
 src/
 ├── main.rs              # Entry point, event loop, action dispatch
+├── config/
+│   └── mod.rs           # User config loading (XDG on Unix, %APPDATA% on Windows)
 ├── app.rs               # Application state (App struct, InputMode, etc.)
 ├── error.rs             # Error types (TuicrError enum)
+├── theme/
+│   └── mod.rs           # Theme palette definitions + CLI theme parsing/resolution
 │
 ├── vcs/                 # VCS abstraction layer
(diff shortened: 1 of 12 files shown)


agavra (approved):
Skimmed the code and tested it locally with all the themes as well as the config (with invalid key/value as well). This is really neat, I've already set my default theme to `catppuccin-macchiato` 😻

PR #166, opened 2026-02-10
fix(config): ignore unknown keys while preserving known settings

mikecfisher:
Hey, I put together a follow up for the config parsing behavior from #162.

Right now an unknown key can cause us to throw out the whole config file, so I changed the loader to parse known keys and ignore unknown keys with a startup warning.

In this change I removed strict unknown field rejection, parse the config table manually, keep valid known keys like theme, and collect warnings for unknown keys or invalid types. I also wired those config warnings into the existing startup warning flow in main so users still get feedback.

I updated the docs in README and AGENTS to match the new behavior.

I also added tests for mixed known and unknown keys, unknown only configs, invalid theme type, and kept the existing malformed TOML behavior as an error. I ran cargo test and everything passes locally as well as doing some light manual QA.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,7 +95,7 @@ src/
 
 ### Data Flow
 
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thanks for the quick fix! If we start getting more complicated configurations we can pull in the `config` crate which handles a lot of this for us

Issue #107, opened 2026-01-16
add :set tabstop for tab sizing


agavra:
nice! I'm happy with this as is, but if you want to follow it up with supporting something like `:set tabstop=4` (vim style setting the tab size) that would be 🔥

_Originally posted by @agavra in https://github.com/agavra/tuicr/pull/106#discussion_r2700271895_

PR #179, opened 2026-02-14
fix(diff): normalize tabs across parsers and add coverage

0xferrous:


0xferrous:
before:
image
after:
image

0xferrous:
ideally i think we should really just show the whitespace as it is, tab should be visible from the diff as tab, and space as a space, and a tab expansion as a tab expansion, maybe could be done when doing #107

agavra (approved):
I agree that we should be a little more consistnet about our whitespace handling. In the meantime this fix is appreciated :)

PR #180, opened 2026-02-14
 fix(ui): make diff row backgrounds consistent to eol

0xferrous:
after/before
image
image


agavra (approved):
very welcome change, thanks again @0xferrous 🙏

PR #182, opened 2026-02-14
feat: show commit message as reviewable entry for single-commit reviews

thiblahute:
## Summary

- When reviewing a single commit (`tuicr -r <commit>`), the commit message now appears as the first entry in the file list and diff view
- In multi-commit reviews, selecting a single commit in the inline selector also shows its commit message
- The commit message entry supports comments, reviewed status, and folding — just like any code file
- Added `body` field to `CommitInfo` across all VCS backends (git, hg, jj) to capture the full commit message beyond the summary line

## Test plan

- [x] `tuicr -r <single-commit>` — commit message appears as first entry, can be commented on and marked reviewed
- [x] `tuicr -r <commit-range>` — selecting a single commit shows its commit message; selecting a range hides it
- [x] Commit message entry folds when marked reviewed
- [x] Regular file commenting logic unchanged
- [x] `cargo test` — all 214 tests pass

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -5,8 +5,8 @@ use chrono::Utc;
 
 use crate::error::{Result, TuicrError};
 use crate::model::{
-    Comment, CommentType, DiffFile, DiffLine, LineOrigin, LineRange, LineSide, ReviewSession,
-    SessionDiffSource,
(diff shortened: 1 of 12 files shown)


agavra:
Thanks, just got back from a trip. I'll try to debug this tomorrow to figure out why it's not working on my end.

PR #242, opened 2026-04-10
feat(export): filter comment type legend to used types

sebastianst:
## Summary
- The `Comment types:` line in exported markdown now only lists comment types actually used in the current review, instead of always listing all four
- New `export_legend` config option (default: `true`) to suppress the legend line entirely via `export_legend = false` in `config.toml`

## Test plan
- [x] Existing tests updated to verify legend filtering
- [x] New test: legend omitted when `show_legend` is false
- [x] New test: legend only lists used types (default types)
- [x] New test: legend only lists used types (custom types)
- [x] New tests: `export_legend` config parsing
- [x] Full test suite passes (339 tests)
- [x] Manual testing: added comments with subset of types, verified `y` export only lists used types

🤖 *Generated by Claude Code*

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -323,6 +323,8 @@ pub struct App {
     pub saved_inline_selection: Option<(usize, usize)>,
     /// Path filter for scoping diff to a specific file or directory
     pub path_filter: Option<String>,
+    /// Whether to include the "Comment types:" legend line in export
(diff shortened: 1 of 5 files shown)


agavra (approved):
Thanks @sebastianst! This makes sense to me

PR #250, opened 2026-04-22
feat: unset reviewed if file has changed, fixes #191

### LINKED ISSUE #191: feat: track "latest completed review"
My workflow is typically:
* Prompt a plan, claude does output
* Review 1 -> Make changes
* Review 2 (only those changes)

To make review 2 (and 3, 4 etc.) easier, I'd really like to track the commit for the "latest completed review" and easily review from there.

I've made a thin python wrapper, but I think this could make sense in the core tool as well. What do you think? :-)

a-h:
Fixes #191

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -732,7 +732,7 @@ impl App {
     ) -> Result<Self> {
         // Ensure all diff files are registered in the session
         for file in &diff_files {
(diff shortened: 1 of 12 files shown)


agavra (approved):
Thanks again @a-h this LGTM and the behavior makes sense (as discussed previously)

Issue #14, opened 2026-01-08
improve vim-like bindings


agavra:
the vim bindings are decent but I think we should spend some time really thinking them through and coming up with a better binding experience, particularly the comment editing experience (ctrl+s to save is a little janky)

agavra:
@icorbrey go for it! I've fixed some of the stuff and made the experience generally better but if there were any bindings you found to be awkward I'm always open to improving it. In general my inspiration are vanilla VIM bindings + NERDTree bindings for panel navigation.

PR #255, opened 2026-05-04
enable view scrolling via ctrl-{e,y}

natebrennand:
In service of #14, enables scrolling the view a single line at a time regardless of present cursor location.


agavra (approved):
🫡 this is great, TIL about these key bindings haha. I've always used ctrl-d ctrl-u

Issue #29, opened 2026-01-09
improve comments on side-by-side view


agavra:
Thanks @krisajenkins this is awesome!

Commenting is a little awkward though. If this gets complex we can leave it for a follow-up, but I wonder if it makes sense to have `tab` cycle through the now "three" panes (file browser, left side view, right side view) so you can explicit comment on "new" code instead of leaving a comment floating in mid air.

We might be able to leverage the changes made in #20 that add a "new/old" enum in the comment storage format.

EDIT: I suspect this is going to be complicated, let's leave that for a follow up and keep current behavior for now.

<img width="1473" height="964" alt="image" src="https://github.com/user-attachments/assets/2dfaeba5-27b3-4348-9e44-dcdb83faac22" />

_Originally posted by @agavra in https://github.com/agavra/tuicr/issues/28#issuecomment-3729651021_

PR #256, opened 2026-05-04
add cursor highlight option

natebrennand:
This PR introduces a config option to highlight the entire line the cursor is on. This is like the `cursorline` feature in vim (in service of #14).

I found this was very necessary to make the diff view viable while working with lots of new file content. I guess this relates to #29 too.

https://github.com/user-attachments/assets/9a9f1fec-f81e-4be1-a0b9-b2bb8e5de5f5

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -794,7 +794,7 @@ impl App {
             pending_confirm: None,
             supports_keyboard_enhancement: false,
             show_file_list: true,
-            cursor_line_highlight: false,
+            cursor_line_highlight: true,
             file_list_area: None,
             diff_area: None,
             expanded_dirs: HashSet::new(),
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -786,6 +794,7 @@ impl App {
             pending_confirm: None,
             supports_keyboard_enhancement: false,
             show_file_list: true,
+            cursor_line_highlight: false,
```
agavra: gave this a spin and I'm honestly happy making it the default. love it

src/config/mod.rs
```diff
@@ -27,6 +27,7 @@ pub struct AppConfig {
     pub diff_view: Option<String>,
     pub wrap: Option<bool>,
     pub export_legend: Option<bool>,
+    pub cursor_line: Option<bool>,
```
agavra, line 30: let's also add this to the README since that's the only config reference at the moment

agavra (approved):
LGTM, this is great @natebrennand - take it or leave it on the default, but we should update the README

natebrennand:
> LGTM, this is great @natebrennand - take it or leave it on the default, but we should update the README

Thanks!
Happy to make it the default. Will push some tweaks and update the README.

PR #259, opened 2026-05-05
feat: stage file

OrenSayag:
# Summary

- Provide 's' hotkey to staged current file from unstaged diff view.
- Toasts error message when pressing the hotkey on staged diff view
image
- Auto reloads diff when staging

### DIFF
--- a/src/input/keybindings.rs
+++ b/src/input/keybindings.rs
@@ -90,9 +90,6 @@ pub enum Action {
     CollapseAll,
     SelectFileFull,
 
-    // Staging
-    StageFile,
-
     // No-op
     None,
 }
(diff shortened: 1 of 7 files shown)


src/input/keybindings.rs
```diff
@@ -168,6 +171,9 @@ fn map_normal_mode(key: KeyEvent) -> Action {
         (KeyCode::Char('o'), KeyModifiers::NONE) => Action::ExpandAll,
         (KeyCode::Char('O'), _) => Action::CollapseAll,
 
+        // Staging
+        (KeyCode::Char('s'), KeyModifiers::NONE) => Action::StageFile,
```
agavra: I'd rather not overload standard VIM keybindings, let's use a `:stage` command instead

agavra (commented):
Thanks @OrenSayag this is a cool feature, left a comment inline

OrenSayag:
@agavra Changed to `:stage` and made it stage all reviewed files instead of current file - what do you think?

agavra (approved):
I like that behavior, makes sense to me to stage in bulk.

PR #260, opened 2026-05-06
feat(input): support Alt+Backspace to delete previous word

ptoussai:
*Written by Claude*

## What

Adds `Alt+Backspace` (Option+Delete on macOS) as a 'delete previous word' binding alongside the existing `Ctrl+W`. Applied to comment, command (`:`), and search (`/`) input modes for consistency.

## Why

`Alt+Backspace` is the readline standard on Linux and the system-wide "delete word" shortcut on macOS. It was the only common 'delete word' chord that didn't work in tuicr — `Ctrl+W` and `Cmd+Backspace` (the latter from #87) already did.

## Notes

- `Cmd+Backspace` continues to map to `DeleteWord` in comment mode (unchanged from #87). Most macOS terminals translate the chord to plain backspaces before the app sees it, so the visible behavior is 'clear line' for most users — but on terminals that forward the modifier (e.g. Kitty keyboard protocol), it does delete-word, matching prior behavior.
- Added a small unit test asserting `Alt+Backspace → DeleteWord` in all three text input modes.
- README's Comment Mode table updated with the new binding.

## Test

```
cargo test
cargo fmt --check
```

Dogfooded: used the new binding while drafting comments on the very diff that introduces it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ dist/
 | `Enter` / `Ctrl-Enter` / `Ctrl-s` | Save comment |
 | `Shift-Enter` / `Ctrl-j` | Insert newline |
 | `←` / `→` | Move cursor |
-| `Ctrl-w` | Delete word |
+| `Ctrl-w` / `Alt-Backspace` / `Cmd-Backspace` | Delete word |
 | `Ctrl-u` | Clear line |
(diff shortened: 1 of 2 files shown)


agavra (approved):
🥳 awesome stuff, thanks @ptoussai

PR #261, opened 2026-05-06
feat(mouse): wheel + clicks behind opt-in 'mouse = true' (closes #140)

### LINKED ISSUE #140: mouse click makes it impossible to copy code
we should figure out how to allow mouse click to change focused panels but still allow copying code lines, perhaps focus change should be cmd+click or something like that. alternatively I'd prefer we just requier tab to cycle panels instead of a mouse click

cc @YPares

ptoussai:
## What

Re-enables mouse support, gated behind an opt-in config flag so users who hit #140 can keep current behavior unchanged.

Add to your config to turn it on:

```toml
mouse = true
```

When enabled:
- **Wheel** scrolls the panel under the cursor (file list, diff, help popup) by 3 lines, without moving the cursor.
- **Click on a file** jumps to it (lazygit-style).
- **Click on a directory** expands or collapses it.
- **Click on a diff line** positions the cursor on that line.

Default is off. With the flag off, `EnableMouseCapture` is never called and the binary behaves identically to today's main.

## Why

History recap:
- #117 added wheel + click-to-focus.
- #140 reported clicks broke text-copy in the terminal.
- #152 reverted everything.

The root cause of #140 is `EnableMouseCapture` itself: once it is on, the terminal stops forwarding drag-to-select natively, regardless of which mouse events the app handles. There is no in-app workaround that gives you both. lazygit lives with the same trade-off and gates it behind `gui.mouseEvents`. This PR follows the same pattern.

When the flag is on and you want to copy text from the terminal, hold your terminal's bypass modifier while dragging (commonly **Shift** or **Option/Alt**, depending on the terminal).

## Scope

- Wheel scroll (file list, diff, help popup).
- Single left-click on file list (file = jump, directory = toggle) and on diff content (positions cursor).
- Config flag `mouse` (boolean, default `false`), parsed and validated like the other booleans.

Out of scope (future PRs):
- Drag-to-select with cell precision plus a copy command (`:y` family).
- Double/triple-click and middle-click paste.

## Test

`cargo test --release` -> 355 pass (3 new for the mouse config flag, 1 guard test that no key produces mouse-scroll actions).

Manual:
- Mouse off (default): wheel does nothing in the TUI, terminal native selection works. Confirms no regression vs current main.
- Mouse on: wheel scrolls without cursor jump, click on file jumps, click on directory toggles, click on diff line positions cursor. Hold the bypass modifier while dragging to copy.
- Quit and panic both restore the terminal cleanly.

Closes #140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,23 @@ dist/
 !Cargo.lock
 ```
 
+### Mouse
+
+Mouse support is **opt-in**. Enable it in your config:
+
+```toml
+mouse = true
+```
+
+| Action | Effect |
+|--------|--------|
+| Wheel up/down | Scroll the panel under the cursor (file list, diff, or help popup) without moving the cursor line |
+| Click on a file | Jump to that file (lazygit-style) |
+| Click on a directory | Expand or collapse it |
+| Click on a diff line | Position the cursor on that line |
+
+When mouse capture is on, the terminal stops handling drag-to-select natively. To copy text, hold your terminal's bypass modifier while dragging (commonly **Shift** or **Option/Alt**, depending on the terminal). Check your terminal's docs if neither works.
+
 ### Keybindings
 
 #### Navigation
--- a/src/app.rs
+++ b/src/app.rs
@@ -141,6 +141,24 @@ pub enum FindSourceLineResult {
     NotFound,
 }
 
+pub fn annotation_file_idx(annotation: &AnnotatedLine) -> Option<usize> {
+    match annotation {
+        AnnotatedLine::FileHeader { file_idx }
+        | AnnotatedLine::FileComment { file_idx, .. }
+        | AnnotatedLine::HunkHeader { file_idx, .. }
+        | AnnotatedLine::DiffLine { file_idx, .. }
+        | AnnotatedLine::SideBySideLine { file_idx, .. }
+        | AnnotatedLine::LineComment { file_idx, .. }
+        | AnnotatedLine::BinaryOrEmpty { file_idx } => Some(*file_idx),
+        AnnotatedLine::ReviewCommentsHeader
+        | AnnotatedLine::ReviewComment { .. }
+        | AnnotatedLine::Expander { .. }
+        | AnnotatedLine::HiddenLines { .. }
+        | AnnotatedLine::ExpandedContext { .. }
+        | AnnotatedLine::Spacing => None,
+    }
+}
+
 /// Search `line_annotations` for the annotation whose `new_lineno` best matches
 /// `target_lineno` within the file identified by `current_file`.
 pub fn find_source_line(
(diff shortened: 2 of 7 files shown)


agavra:
thanks for the patch @ptoussai but this changes the behavior of the mouse to be incompatible with what VIM does. The current behavior before this patch is preferred (scrolling does not scroll the viewport, it scrolls the cursor).

YPares:
@agavra Allow me to question the need to stick to vim's conventions ;) (most of which date back to _before_ we had scroll wheels)
And I speak as a modal editor user, so I harbour no ill will.

agavra:
😆 fair point, my line of thought is twofold:

1. it makes decisions a lot easier to make, and we can move fast without bikeshedding UX experience
2. people who come from vim will feel familiar with the UX for tuicr

I think if something came along that doesn't make much sense for us and departing from vim is really helpful we should be open to it, but to me the default is to match vim

ptoussai:
Thanks for the feedback @agavra!

On vim's mouse behavior with :set mouse=a: the wheel scrolls the viewport (cursor stays put), click moves the cursor, drag enters visual mode, and double/triple-click select word/line. What vim doesn't do is auto-copy on release.
So the wheel-scrolls-viewport part of this PR actually matches vim. The pre-patch behavior (wheel = cursor) is the unintentional opposite, It's the fallback you get when nothing captures the mouse.
I believe it's natural for most users

Drag-to-auto-copy is a departure: the vim idiom would be drag, then visual, then y, but y is already the export-comments shortcut and v/V is reserved for line-range comments.

More importantly, this PR is really about getting the mouse-capture infrastructure in place. With it on, future PRs (file/line click-to-select, drag-as-visual, etc.) become straightforward.

ptoussai:
Note: `:set mouse=a` is the vim default, cf. https://github.com/vim/vim/blob/master/runtime/defaults.vim#L76
But it's not the case on all default vim installations. On macOS for example it's not set.

agavra:
Got it, I didn't know that! I just take my default vim experience as gospel. Let me review the PR with that in mind.

ptoussai:
I reduced the scope of this PR since the original push, so I've rewritten the title and description above.

What changed:
- Removed the in-app drag-to-copy / auto-copy. The approach worked but the highlight had an off-by-one and cell-precise selection needed more work than belonged in this PR.
- Reduced to mouse infrastructure only: wheel + single clicks (file = jump, directory = toggle, diff line = position cursor).
- Gated everything behind a new `mouse = true` config flag, default off. Closes #140 by making the trade-off opt-in. We can revisit the default with a proper copy in place.

Deferred to a follow-up: a proper copy command (`:y` family) plus cell-precise drag selection.

agavra:
Ah! I was just typing my update that I didn't love the drag-to-copy default since it messed with my clipboard history. The rest looked good, but let me take another look at the new scope.

agavra (approved):
New scope and behavior LGTM. Thanks @ptoussai when we figure out a good default mouse UX we can take it from there

ptoussai:
Thanks for the review and the quick merge!

PR #262, opened 2026-05-07
feat(mouse): drag-to-select with cell-precise copy and visual range comments

ptoussai:
Drag with the left mouse button in the diff to highlight a cell-precise range. Press `y` to copy the selected source content (no line numbers, no `+`/`-` markers); press `c` (or Enter) to open a range comment for the line span.

Keyboard `V` and mouse drag share one pipeline: same data model, same renderer, same actions. The selection is anchored to content (annotation index + char offset) so it survives scroll. When the cursor scrolls off-viewport the selection clears (helix-style). In side-by-side, drag is pinned to the pane the press started in.

Hunk headers (`@@ -38,20 +38,27 @@ ...`) and file headers (`src/foo.rs [M]`) participate as atomic rows: they paint a full-row highlight when in the selection and contribute their full text (decoration stripped for file headers) to the copy, giving the pasted block useful location context.

Behind the existing `mouse = true` opt-in flag from #261.

## Try it

- Drag across a few diff lines, press `y` to copy.
- Drag across a hunk header into a code block, press `y` - the `@@` line is included verbatim.
- `V` keyboard mode still works exactly as before for line-based range comments.
- Drag in side-by-side: the selection sticks to whichever pane (Old or New) you pressed in.

## Notes

- All 359 tests pass; clippy clean.
- No new dependencies.
- Native terminal text selection (Option/Shift drag bypass) is unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,8 +214,9 @@ mouse = true
 | Click on a file | Jump to that file (lazygit-style) |
 | Click on a directory | Expand or collapse it |
 | Click on a diff line | Position the cursor on that line |
+| Drag in diff | Highlight a range; press `y` to copy the selected source lines |
 
-When mouse capture is on, the terminal stops handling drag-to-select natively. To copy text, hold your terminal's bypass modifier while dragging (commonly **Shift** or **Option/Alt**, depending on the terminal). Check your terminal's docs if neither works.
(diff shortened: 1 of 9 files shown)


agavra (approved):
nice, I tested this out and I think the behavior makes sense to me. the overloading of `y` for clipboard export of the review was probably a bad choice early on, but one we're stuck to. I'm OK with this behavior for now. Maybe next major release (1.0) we can rework that

PR #264, opened 2026-05-07
feat(theme): github-light, github-dark, and --transparent flag

ptoussai:
Adds two GitHub-themed presets and a `--transparent` flag.

## Themes

- `--theme github-light` matches github.com's light diff palette (Primer light tokens). Syntax via `InspiredGitHub`.
- `--theme github-dark` matches github.com's dark palette. Syntax via `OneHalfDark` (closest bundled fit; no GitHub Dark theme is shipped with `two-face`).

## --transparent

Clears the panel background so the terminal's own background shows through. Diff add/del backgrounds, status bar, cursor line, and list selection still paint, so colored content stays visible.

```sh
tuicr --theme github-dark --transparent
```

Also available as `transparent_background = true` in the config file.

355 tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,11 @@ Detection order: Jujutsu → Git → Mercurial. Jujutsu is tried first because j
 | Flag | Description |
 |------|-------------|
 | `-r` / `--revisions <REVSET>` | Commit range/Revision set to review. Exact syntax depends on VCS backend (Git, JJ, Hg) |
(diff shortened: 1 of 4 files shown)


ptoussai:
This is how `tuicr --theme github-light  --transparent` looks on ghosty with `theme = Catppuccin Latte`

SCR-20260507-qlaz

agavra (approved):
❤️ I love it! Thanks @ptoussai

PR #266, opened 2026-05-07
feat(mouse): wheel + click on commit list (full-screen and inline)

ptoussai:
Wheel + click on the two commit lists that mouse hadn't reached yet.

- Full-screen commit picker (`InputMode::CommitSelect`): wheel moves the cursor, click positions the cursor and toggles selection (or expands the row when on `... show more commits ...`).
- Inline commit selector at the top of the diff in multi-commit reviews: wheel and click route the same way; click also focuses the panel and dispatches the same action `Enter` does (toggle + reload).

Both lists are short and selection-oriented, so wheel ticks move the cursor rather than viewport-scrolling away from it - matches `j`/`k`.

```sh
cargo install --path .
tuicr  # try clicking commits in the picker
```

359 tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,10 +212,11 @@ mouse = true
 
 | Action | Effect |
 |--------|--------|
-| Wheel up/down | Scroll the panel under the cursor (file list, diff, or help popup) without moving the cursor line |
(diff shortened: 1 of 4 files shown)


agavra (approved):
@ptoussai you're on fire, thanks again!

PR #267, opened 2026-05-07
fix(ui): cursor-line highlight covers +/- code in unified diff

ptoussai:
In unified diff view, the cursor-line highlight only covered the gutter (indicator + line number), not the actual code on `+`/`-` lines.

Cause: `paint_unified_diff_row_backgrounds` painted `cursor_line_bg` *before* the Paragraph rendered. Spans on `+`/`-` lines carry their own `diff_add_bg` / `diff_del_bg` (and syntax-highlighted spans bring their own bg), which then overwrote the cursor highlight on the content cells. Side-by-side already gets this right because it post-paints.

Fix: move the cursor highlight to a post-paint after `render_widget`. Both passes go through one closure-based helper so the wrap-aware row iteration isn't duplicated.

```sh
cargo install --path .
tuicr  # diff view, cursor on a +/- line — whole row is now lit
```

359 tests pass.

### DIFF
--- a/src/ui/app_layout.rs
+++ b/src/ui/app_layout.rs
@@ -1434,8 +1434,7 @@ fn render_unified_diff(frame: &mut Frame, app: &mut App, area: Rect) {
             app.diff_state.wrap_lines,
             inner.width as usize,
             |idx, _line| {
-                is_line_highlighted(app, idx)
-                    .then(|| Style::default().bg(app.theme.cursor_line_bg))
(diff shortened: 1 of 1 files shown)


agavra (approved):
I assumed this may have been intentional by @natebrennand but I think I prefer this experience so let's go with your patch :)

agavra:
PS I'll cut a release soon-ish after all these changes are in

ptoussai:
I also wondered but it seems like a bug to me ... and it also looked strange and inconsistent between views.

PR #268, opened 2026-05-07
feat(ui): auto-clear status messages after a TTL

ptoussai:
Status messages used to stay on screen until something else overwrote them. Now they auto-clear:

- Info: 3s
- Warning: 5s
- Error: sticky (failures should stay visible until acknowledged)

The dirty-quit prompt ("Unsaved changes. Press q again to quit.") is the one warning that needs to outlive the timer because it's tied to `quit_warned`. It uses a new `set_sticky_warning` helper that sets `expires_at = None`.

Main loop runs `app.clear_expired_message()` once per iteration, before the draw, so an expired message disappears within ~100ms (the poll interval).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
+use std::time::{Duration, Instant};
 
 use chrono::Utc;
 use ratatui::style::Color;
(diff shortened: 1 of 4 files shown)


agavra (approved):
tbd if I like this, I don't see too much harm in leaving messages around in case you missed it, but we can merge it for now and I'll see if I miss the old behavior

ptoussai:
It could be tuned and some messages could be sticky. We could also decide to have all warnings as sticky instead.

Issue #214, opened 2026-03-18
Light mode themes should not have background highlighting


guineveresaenger:
I would like to have a light mode theme that _doesn't_ highlight the entire background text in color. I'd prefer to see a simple git diff display without any text highlighting.

None of the light modes enable you to read the text, because the lack of contrast unfortunately just  washes out the words, with the possible exception of gruvbox, and that color's a real strain on the eyes.

I know all screens render things differently so I'm not sure how much these screenshots help:

ayu-light:
<img width="1126" height="296" alt="Image" src="https://github.com/user-attachments/assets/dbe69496-9426-42c8-a805-3073e363fdca" />

catppuccin-frappe:
<img width="1230" height="326" alt="Image" src="https://github.com/user-attachments/assets/d49b282c-3cff-4724-9739-4771aa01e09a" />

catppuccin-latte:

<img width="528" height="157" alt="Image" src="https://github.com/user-attachments/assets/11d86967-dba3-4758-b8d3-e297cf877e1a" />

default light:

<img width="744" height="287" alt="Image" src="https://github.com/user-attachments/assets/d09e5c84-b7a2-423b-a36f-d7fce7fc0838" />

gruvbox-light (workable but really barely):

<img width="1151" height="234" alt="Image" src="https://github.com/user-attachments/assets/b4c5b844-e070-4aa6-91af-3a4123c60a90" />

agavra:
I am curious to know which terminal you are using, though also your suggestion has merit. I don't think I'll be able to craft a good color scheme for you without also being able to reproduce the problem, but do feel free to submit one and I'll merge it in :)

PR #269, opened 2026-05-07
feat(theme): default to transparent panel background

ptoussai:
Breaking change, opening for discussion.

## Why

Popular modern TUIs like lazygit and Claude Code show the terminal background through their panels. tuicr currently paints the theme's `panel_bg` instead. Flipping that, so terminal customizations (images, blur, colors) show through out of the box. Also looks nicer, especially for the default light/dark themes. The light theme's bg in particular is a bit ugly, imo :)

## What changes

- Default: transparent panel backgrounds.
- Opt out: `transparent_background = false` in config.
- `--transparent` flag removed (unreleased, only on main).

## Alternatives considered

- Bake into theme: themes are hardcoded in source, so no escape hatch for users who want a palette with transparent panels.
- Suffix on `--theme` (`catppuccin-mocha:transparent`): new parser grammar for one boolean.
- Conditional default (transparent only when no explicit theme): harder to reason about than a flat config key.

## Compatibility

- `transparent_background = true` in config: no change.
- No setting: panels go transparent on next launch.
- Want the painted bg: add `transparent_background = false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,6 @@ Detection order: Jujutsu → Git → Mercurial. Jujutsu is tried first because j
 | `--appearance <MODE>` | Appearance mode for default theme (`dark`, `light`, `system`) |
 | `--stdout` | Output to stdout instead of clipboard when exporting |
 | `--no-update-check` | Skip checking for updates on startup |
-| `--transparent` | Use the terminal's background color |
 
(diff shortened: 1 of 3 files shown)


agavra:
I wonder if this will fix #214 - I've never been able to reproduce it. Let's go for it.

agavra (approved):

PR #270, opened 2026-05-07
feat(mouse): on by default

ptoussai:
## Why

Most modern TUIs ship mouse on by default. The default UX is better with it, especially wheel-scrolling.

Mouse was opt-in (#261) because capture broke native copy (#140). #262 replaced that with in-app drag-to-select that's actually nicer than native: it copies clean source across multiple lines and includes file headers.

Native terminal selection still works via the Shift / Option bypass already documented in the README.

## What changes

- Default: `mouse = true`.
- Opt out: `mouse = false` in config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,10 +204,10 @@ dist/
 
 ### Mouse
 
-Mouse support is **opt-in**. Enable it in your config:
+Mouse support is **on by default**. To disable, set in your config:
 
 ```toml
-mouse = true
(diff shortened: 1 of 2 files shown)


agavra (approved):
I've been using `mouse = true` for the last day and it's definitely been a good UX. 🔥 thanks for all the work on it @ptoussai I'll cut a release tonight!

PR #273, opened 2026-05-11
fix(syntax): full-file context for container-grammar diffs (vue, svelte, astro, mdx)

ptoussai:
Vue diffs (and Svelte / Astro / MDX / PHP / ERB-family templates, same root cause) render unhighlighted today. syntect's container grammars (`text.html.vue`, `text.html.php`, etc.) only enter their nested JS / HTML / CSS / template scope after seeing the opening `<template>`, `<script>`, `<style>`, `<?php`, or `<% %>` tag. The current code highlights only the hunk slice, so a hunk that lives inside `<script>` (or after `<?php`) never sees the opener, stays in its outer scope, and every span comes back with the theme's default foreground.

## Fix

An extension gate (`vue`, `svelte`, `astro`, `mdx`, `php`, `erb`, `eex`, `heex`) decides whether to pay the cost. Other files keep per-hunk highlighting unchanged. For matched files each backend obtains the full old and new content, re-highlights end-to-end, then re-keys each diff line by its 1-based lineno.

The shape per backend is the same: a "batched fetch" closure that returns full file content for a given list of paths at a given revision. Only container-grammar files are ever passed to it, so non-container files in the diff pay nothing extra.

- **git**: `git2` blob lookup in-process (`Tree::get_path` + `find_blob`; index / workdir for the staged, unstaged, and working-tree cases). Zero overhead.
- **hg**: one `hg cat -r REV --template '\n{BATCH_BOUNDARY}\n{path}\n{data}' file1 file2 ...` per side. hg cat is Python-startup-bound (~280 ms regardless of file count), so batching all paths into one call is essentially free.
- **jj**: one `jj file show -r REV -T '"\n{BATCH_BOUNDARY}\n" ++ path ++ "\n"' file1 file2 ...` per side. Sub-20 ms total.

Shared in `vcs/mod.rs`: `enhance_with_full_file_highlight` (closure-based, used by all three backends), `apply_container_full_file_highlight` (collects container paths, runs the backend's batched fetch, calls the enhance pass), and `parse_batched_files` for the boundary-delimited output. The git path uses its own in-process blob helpers; hg/jj plug in their batched-fetch subprocess wrappers.

Files larger than 1 MiB or containing null bytes skip and stay monochrome.

A regression test was added for each backend: build a real repo with a `.vue` file, edit a line inside `<script setup>`, assert the changed lines have multiple distinct foreground colors.

## Why not just inflate diff context?

An earlier draft of this PR bumped hg/jj to `-U 1000000` / `--context 1000000` and reconstructed file content from the resulting hunks. That avoided extra subprocesses but inflated every file's hunk to the whole file, so per-hunk syntax highlighting on non-container files (TS, CSS, Rust, etc.) ran over full-file content via syntect. On a worst-case mixed diff (50 Vue + 150 TS modified) that cost hg 5.60 s and jj 3.62 s. The current batched-fetch approach drops them to 4.58 s and 1.85 s by keeping non-container files on the cheap small-hunk path.

## Cost model

For matched files we pay syntect on the full file, twice (old + new sides). syntect runs at roughly 27 us/line on Vue / 37 us/line on TS, so the total scales linearly with `(container lines in diff) * 2`.

- **Realistic PR** (1-5 Vue files, ~200 lines each): ~50 ms of extra highlighting. Imperceptible.
- **Pathological** (50 Vue files at ~620 lines each, all modified in one review): ~1.6 s of extra highlighting. The 62k-line figure is genuinely the work syntect has to do to produce correct spans across template/script/style boundaries; there is no shortcut inside syntect itself.

That worst case is what it looks like to review an entire component library's rewrite in a single pass. It's not what tuicr's hot path looks like. If it ever becomes a real problem, file-level rayon parallelism in `enhance_with_full_file_highlight` is a straightforward follow-up (independent per file, syntect's `SyntaxSet` and `Theme` are `Sync`).

### DIFF
--- a/src/syntax/mod.rs
+++ b/src/syntax/mod.rs
@@ -5,10 +5,42 @@ use two_face::theme::EmbeddedThemeName;
 use crate::model::diff_types::LineOrigin;
 
 /// A single line of highlighted spans (style + text pairs).
-type HighlightedSpans = Vec<(Style, String)>;
+pub(crate) type HighlightedSpans = Vec<(Style, String)>;
 
 /// Per-line highlight results for a file: `Some` if the line was highlighted, `None` on failure.
-type HighlightedLines = Vec<Option<HighlightedSpans>>;
+pub(crate) type HighlightedLines = Vec<Option<HighlightedSpans>>;
+
+/// Whether `path` belongs to a "container" syntax that embeds other languages
+/// and so needs the full file (not just a hunk slice) in scope before its
+/// nested grammars activate.
+///
+/// Sublime-syntax grammars start in a `main` context entered at the top of the
+/// file. For Vue, syntect stays in `text.html.vue`'s outer scope until it
+/// sees `<template>`, `<script>`, or `<style>`. Same shape for Svelte / Astro
+/// / MDX (fenced code blocks) / PHP / ERB-family templates: anything inside a
+/// nested block in a hunk that doesn't include the opening tag will fall back
+/// to the theme's default foreground.
+///
+/// Container extensions kept narrow. `html` and `md` are intentionally
+/// omitted: the vast majority of changes in those files are outside any
+/// embedded-language block, so paying the full-file cost on every diff would
+/// be a net regression. Add only when an extension is overwhelmingly used as
+/// a container (i.e. most hunks live inside a nested grammar).
+pub(crate) fn needs_full_file_highlight(path: &Path) -> bool {
+    matches!(
+        path.extension().and_then(|e| e.to_str()),
+        Some(
+            "vue"     // text.html.vue: HTML + JS/TS + CSS blocks
+            | "svelte" // source.svelte: HTML + JS/TS + CSS blocks
+            | "astro"  // source.astro: frontmatter + HTML + JS
+            | "mdx"    // text.html.markdown with JSX inside
+            | "php"    // text.html.php: outer HTML, switches at <?php
+            | "erb"    // text.html.erb: outer HTML, switches at <% %>
+            | "eex"    // text.html.elixir: same shape as erb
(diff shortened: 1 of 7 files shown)


agavra (approved):
Tested locally and the approach LGTM, though it did help me realize that we've significantly regressed in performance on large PRs since 0.11.0. I'll open a separate ticket for that and see if I can get the agents to speed it up.

ptoussai:
Quick check on the regression since v0.11.0 since I had the bench harness handy from #280. Couldn't repro it.

Method: temp git repo, N TypeScript files at M lines each, modified, then `App::new(-r HEAD~1..HEAD)` + a `terminal.draw(|f| ui::render(f, &mut app))` loop on a `TestBackend` 200x60.

| workload | metric | main | v0.11.0 |
|---|---|---:|---:|
| 30 files x 300 lines | App::new | 1.09 s | 1.09 s |
|  | render | 11.7 ms | 11.5 ms |
|  | scroll+render | 11.6 ms | 11.7 ms |
| 100 files x 500 lines | App::new | 5.91 s | 5.90 s |
|  | render | 62.1 ms | 64.6 ms |
|  | scroll+render | 61.8 ms | 64.2 ms |

Within run-to-run noise across diff load, App construction, and render-while-scrolling. Whatever you noticed is likely in a path the synthetic bench doesn't exercise (specific repo content, event loop, commit-select screen, etc.). Dropping the numbers here in case they're useful when you open the ticket.

agavra:
@ptoussai ah glad you had it available, I realized it was because I was being stupid and compared a release build with a debug build 😆 (see #278)

ptoussai:
Thanks for the review and merge!

PR #277, opened 2026-05-11
provide option for cursor offsets

natebrennand:
This PR adds support for a `scroll_offset` config option that maintains a buffer of extra lines between the cursor and the top/bottom of the view. I find this useful to ensure that there is some context visible before/after wherever I have my cursor, otherwise I will need to navigate too far then scroll back.

This demonstrates behavior when using `scroll_offset = 4` in my config:

https://github.com/user-attachments/assets/cb199433-1beb-4cdf-90a7-e35181a8dc42

I did my best to match the behavior of vim in how this feature works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ diff_view = "side-by-side"
 wrap = true
 cursor_line = false
 transparent_background = false
+scroll_offset = 5
 
 comment_types = [
   { id = "note", label = "question", definition = "ask for clarification", color = "yellow" },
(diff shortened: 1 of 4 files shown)


agavra (approved):
Love it, I'll probably set this as the default for me as well. Thanks  @natebrennand 

Let's just update the README so it's discoverable.

r-vdp:
In neovim I keep my cursor always in the middle, so basically it's like the viewport scrolls underneath the cursor instead of the cursor moving inside the viewport (only for the vertical direction, obviously).

I might have a go at implementing that here too. The main avantage is that that automatically adapts to the size of the window, whereas with the offset you are limited to small offsets if you don't want to run into issues with small windows.

PR #280, opened 2026-05-11
perf(syntax): parallelize full-file highlighting across files

ptoussai:
Follow-up to #273. That PR re-highlights `.vue` / `.svelte` / `.astro` / `.mdx` / `.php` / `.erb`-family files using full-file content (syntect needs the opening container tag in scope), so highlighting scales linearly with the number of such files in the diff. 25 modified `.vue` files spends ~200 ms in syntect on diff load.

## Fix

Refactor `enhance_with_full_file_highlight` into three phases:

1. Fetch (serial): collect content via the backend's fetcher closure. Stays serial because git's closure holds a `!Send` `git2::Repository`.
2. Highlight (parallel): syntect's `SyntaxSet` and `Theme` are `Sync` and each file is independent. Chunked across `std::thread::available_parallelism()` via `std::thread::scope`. No new dependency.
3. Apply spans (serial): write spans back into `&mut files`.

## Numbers

Wall-clock on 8-core macOS, 200-line synthetic Vue files:

| files | serial    | parallel  | speedup |
|------:|----------:|----------:|--------:|
| 1     | 8.79 ms   | 8.86 ms   | 0.99x   |
| 5     | 43.98 ms  | 12.16 ms  | 3.62x   |
| 12    | 106.07 ms | 17.98 ms  | 5.90x   |
| 25    | 219.78 ms | 31.47 ms  | 6.98x   |
| 50    | 441.48 ms | 60.05 ms  | 7.35x   |

N=1 stays on the serial path; the threshold avoids paying thread-setup cost when there's nothing to parallelize.

### DIFF
--- a/src/vcs/mod.rs
+++ b/src/vcs/mod.rs
@@ -145,6 +145,12 @@ const MAX_HIGHLIGHT_FILE_BYTES: usize = 1024 * 1024;
 /// available, every diff line on that side is replaced with the span at its
 /// 1-based lineno from the full-file highlight. Lines whose side could not be
 /// fetched keep whatever the parser already assigned.
+///
+/// Runs in three phases: fetch (serial, since fetch closures may close over
+/// `!Send` state such as `git2::Repository`), highlight (parallel via
+/// `std::thread::scope`, since syntect's `SyntaxSet` and `Theme` are `Sync`
+/// and each file's syntect work is independent), and apply spans (serial,
+/// needs `&mut files`).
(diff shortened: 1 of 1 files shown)


agavra (approved):
LGTM and simple enough to be worth merging, but it does make me wonder if we should look in to restructuring the rendering path so we can parallelize all of the loading jobs and not just the full file highlights. That would help loading time for all PRs and we could remove this special-case for full-file highlighting

PR #281, opened 2026-05-12
Fix sparse checkout diff support

### LINKED ISSUE #279: support for sparse git indexes
Many corporate monorepos rely on the use of sparse git indexes to scale, `libgit2` (the underlying dep tuicr uses today) does not support them: https://github.com/libgit2/libgit2/issues/2263

iheanyi:
## Summary
- Add sparse checkout/sparse-index capability detection for Git and route only those repos through Git CLI diff/revision helpers; standard Git repos continue to use the existing libgit2 paths.
- Parse Git-style CLI diffs through the shared streaming parser so sparse checkouts avoid libgit2 sparse-index failures such as unsupported mandatory `sdir` index extensions.
- Preserve sparse checkout untracked-file handling, including binary and large-file cases, while narrowing untracked scans to sparse checkout pathspecs when the sparse patterns are simple cone paths.
- Add startup/profile instrumentation for the critical paths we used to debug large sparse repos, and surface an in-app warning when a sparse checkout does not have `core.untrackedCache` enabled.

## Profiling
This PR keeps temporary `TUICR_PROFILE` support because the sparse checkout path is highly sensitive to repo size and filesystem state. The profiler writes to a file instead of stderr/stdout so it does not corrupt the TUI:

- `TUICR_PROFILE=1` appends to `$TMPDIR/tuicr-profile.log`.
- `TUICR_PROFILE=/path/to/profile.log` writes to that exact file.
- `TUICR_PROFILE_FILE=/path/to/profile.log` overrides the default path when `TUICR_PROFILE` is enabled.

The profiler now covers startup, VCS detection, status probes, diff loading, commit selection, view rebuilds, sparse untracked pathspec selection, untracked enumeration, and cache/fsmonitor metadata. This made it clear that exact untracked enumeration is the remaining expensive operation in very large sparse monorepos; enabling Git's untracked cache reduced the observed untracked scan substantially, but the scan is still real filesystem work.

`core.fsmonitor` can be either a boolean or a custom hook path, so capability detection treats custom non-false fsmonitor values as enabled. `core.untrackedCache` is checked as a boolean because that is the Git knob that makes repeated untracked scans avoid re-reading unchanged directories.

## Git Backend Behavior
The intended branch point is:

- sparse checkout or sparse index: use Git CLI helpers for diff/revision operations that libgit2 cannot reliably handle in sparse mode.
- standard Git repos: keep the previous libgit2 behavior, including the original libgit2 staging path.

That keeps the sparse support additive instead of changing the normal Git backend behavior.

## Test Plan
- `cargo fmt --all --check`
- `cargo clippy --no-default-features -- -D warnings`
- `cargo test`
- `git diff --check`

Fixes https://github.com/agavra/tuicr/issues/279

### DIFF
--- a/src/profile.rs
+++ b/src/profile.rs
@@ -1,0 +1,52 @@
+use std::time::{Duration, Instant};
+
+pub fn enabled() -> bool {
+    std::env::var_os("TUICR_PROFILE").is_some()
+}
+
+pub fn time<T, F>(label: &str, f: F) -> T
+where
+    F: FnOnce() -> T,
+{
+    if !enabled() {
+        return f();
+    }
+
+    let started = Instant::now();
+    let result = f();
+    log(label, started.elapsed());
+    result
+}
+
+pub fn time_with<T, F, M>(label: &str, f: F, metadata: M) -> T
+where
+    F: FnOnce() -> T,
+    M: FnOnce(&T) -> String,
+{
+    if !enabled() {
+        return f();
+    }
+
+    let started = Instant::now();
+    let result = f();
+    log_with(label, started.elapsed(), metadata(&result));
+    result
+}
+
+pub fn log(label: &str, duration: Duration) {
+    log_with(label, duration, String::new());
+}
+
+pub fn log_with(label: &str, duration: Duration, metadata: String) {
+    if enabled() {
+        let suffix = if metadata.is_empty() {
+            String::new()
+        } else {
+            format!(" ({metadata})")
+        };
+        eprintln!(
+            "[tuicr profile] {label}: {:.2}ms{suffix}",
+            duration.as_secs_f64() * 1000.0
+        );
+    }
+}
--- a/src/profile.rs
+++ b/src/profile.rs
@@ -1,7 +1,13 @@
+use std::fs::{File, OpenOptions};
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::{Mutex, OnceLock};
 use std::time::{Duration, Instant};
 
+static PROFILE_LOG: OnceLock<Option<Mutex<File>>> = OnceLock::new();
+
 pub fn enabled() -> bool {
-    std::env::var_os("TUICR_PROFILE").is_some()
+    profile_path().is_some()
 }
 
 pub fn time<T, F>(label: &str, f: F) -> T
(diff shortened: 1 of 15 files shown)


agavra:
@iheanyi is this related to #279?

iheanyi:
@agarva Yep!

iheanyi:
Okay this is ready for review! I added profiling in this branch because it was really helpful in debugging slowness, but more than happy to remove it if you think it adds too much bloat!

agavra:
cc @natebrennand if you can test patch on your repo as well, I know you had similar issues and proposed a similar fix. I'll try to reproduce it with a mock sparse index repo but I don't know if I have something easy to test. I'll see if there's a good OSS one I can clone to test on.

iheanyi:
Funny enough, we work together and we both talked about it at work to try and fix it ourselves. Happy to use whichever patch solves the problem better!

src/profile.rs
```diff
@@ -0,0 +1,100 @@
+use std::fs::{File, OpenOptions};
```
agavra, line 1: I like that we added profiling, but I'd prefer we used the `tracing` crate for this instead of building our own profiling toolchain - collecting spans is probably the preferred way to profile things and it's more flexible with regards to how we do it.

either way, this PR is a bit large so let's save that for another one so it's a bit easier to review 😅

src/vcs/git/mod.rs
```diff
@@ -23,6 +25,75 @@ pub use diff::{
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum GitRepoMode {
+    Standard,
+    SparseCheckout,
+    SparseIndex,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct GitCapabilities {
```
agavra, line 39: instead of modeling `GitCapabilities` within the `git` vcs module, what do you think about modeling it as a different vcs altogether (e.g. `git-cli`)? That will keep the code paths detached instead of having conditional branches throughout the existing git implementation.

---

Here was Codex's take on the suggestion:

I like the direction, with one caveat: I’d model it as a separate **backend implementation**, not a separate user-facing `VcsType`.

So: `GitBackend` for libgit2, `GitCliBackend` for sparse/sparse-index repos, both reporting `VcsType::Git`. Calling it `git-cli` internally is fine, but exposing `VcsType::GitCli` would leak an implementation choice into the status bar, tests, docs, and maybe future persistence semantics. It is still Git from the user’s perspective.

That split would clean up the current design. Right now `GitCapabilities::requires_git_cli()` branches through `diff`, `repository`, `context`, `staging`, and status probing. A separate `GitCliBackend: VcsBackend` would let each method be straight-line CLI code, and the normal `GitBackend` can go back to being libgit2-only.

I’d shape it roughly like this:

```rust
src/vcs/git/
  mod.rs          // detection/shared Git info
  libgit2.rs      // GitBackend
  cli.rs          // GitCliBackend
  common.rs       // shared VcsInfo creation, config detection, command helpers
```

Detection can become:

```rust
if git repo is sparse checkout/sparse index {
    return GitCliBackend::discover();
}
return GitBackend::discover();
```

or `git::discover() -> Result<Box<dyn VcsBackend>>`, so `detect_vcs()` does not care.

A few details I’d keep in mind:

- `GitCapabilities` can shrink into `GitCliFeatures` or `GitRepoFeatures`; it should describe CLI behavior like sparse cones, untracked cache, fsmonitor, warnings, not decide every operation path.
- Keep `VcsType::Git` for both backends. If we want diagnostics, add a separate internal label like `backend_name()` or include it only in profile spans.
- This won’t by itself fix the ignore-filtered status bug, because `.tuicrignore` is applied in `App`, not the VCS backend. The fast status path still needs to respect the same filtered-diff contract or fall back to diff-based probing.
- The binary patch, quotePath, and sparse untracked pathspec fixes become much easier to localize inside `GitCliBackend`.

Net: yes, I’d prefer this. It turns “Git repo capabilities decide branches everywhere” into “VCS detection chooses the right Git implementation once,” which is a cleaner boundary for this codebase.

src/vcs/diff_parser.rs
```diff
@@ -54,8 +78,8 @@ pub fn parse_unified_diff(
             }
 
             // Check if binary - hg uses "Binary file", jj/git use just "Binary"
-            if lines.peek().is_some_and(|l| l.contains("Binary")) {
-                lines.next(); // consume binary message
+            if peek_line(&mut lines)?.is_some_and(|l| l.contains("Binary")) {
```
agavra, line 81: Codex found this one: 
```
  For a tracked binary modification, git diff --binary emits:

  diff --git a/file.bin b/file.bin
  index ...
  GIT binary patch
  literal ...
```

GIT binary patch is lowercase binary, and parse_file_header currently treats it as generic metadata and consumes it before the later check runs. So the review comment is pointing at existing parser code that became insufficient once sparse Git started feeding it git diff --binary output.

agavra (commented):
Thanks @iheanyi happy to have support for sparse indexes :) 

I gave this a quick whirl and confirmed it works on the LLVM project https://github.com/llvm/llvm-project which uses sparse indexes. That said, I think it'd be nice if we:

1. pushed profiling to a different PR, this one is already pretty big to review
2. refactored it to make git-cli a different backend instead of adding lots of conditionals

If you agree (2) is a good strategy, we should stage this in 2 PRs. The first would be to refactor the existing git stuff in preparation for the change, then add the actual git-cli backend implementation.

iheanyi:
@agavra Good call, for (2), how do you envision it being used as a different back-end? Basically just letting it be used everywhere in an all-or-nothing fashion or something else?

agavra:
Yeah I imagine it being an all-or-nothing fashion. I think if you have a local git CLI that you can depend on, and you need it for some features, I think it's fine to just use it for all features. The benefit of shipping with `libgit2` was that you don't need to depend on what's installed for `tuicr` to run (sandbox environments, etc...).

iheanyi:
Makes sense, will fire off an agent to refactor this, standby.

iheanyi:
Superseded by the split replacement PRs:

- #283 for tracing/profile instrumentation
- #284 for the libgit2 backend split
- #285 for the sparse Git CLI backend

The sparse backend PR references the original review threads and keeps sparse Git CLI support as a separate all-or-nothing backend.

iheanyi:
@agavra There you go, let me know what you think about each of those. I have it setup to monitor every PR for your feedback and address it :)

PR #282, opened 2026-05-12
Advance cursor after toggling a commit into the selection

jaredramirez:
Hi! I did not open an issue for this before hand, so if this feature is not desirable, then feel free to close!

The tool [`jjui`](https://github.com/idursun/jjui) has really nice UX when you're selecting sets of commit, it auto moves the cursor to the next commit after selection. This makes it very easy to select a bunch of commits by just pressing `Space` or `Enter` several times in a row (vs having to interleave with `j`, ie `Space`, `j`, `Space`, `j`, and so on).

The semantics are: When the user presses Enter (inline picker) or Space (inline + full-screen picker) to toggle a commit's selection, add the cursor commit to the selection range, *then* move the cursor past the end of the range to the next unselected commit. Lets the user press repeatedly to select a contiguous run of commits without alternating with `j`.

Only "extending" presses advance the cursor. Pressing Enter/Space on an:
* Edge commit deselects the commit
* Middle commit truncate the range while leaving the cursor commit selected

Here's before (must select, then move down, then select, then move down, etc)

https://github.com/user-attachments/assets/b85674d0-f6e5-49c2-bcfb-b53bea044612

Then after (just pressing select repeatedly):

https://github.com/user-attachments/assets/7f3580bb-3cab-496e-bcac-75ad60a13850

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3665,6 +3665,34 @@ impl App {
         }
     }
 
+    /// Toggle the cursor commit's membership in the selection range, then
+    /// (only if the cursor commit was newly added to the selection) move the
+    /// cursor past the end of the range. Lets the user press Enter/Space
+    /// repeatedly to sweep a contiguous run of commits.
+    ///
+    /// Other toggle outcomes leave the cursor in place: edge presses
+    /// (deselect the cursor commit), middle presses (truncate the range
+    /// without unselecting the cursor commit), and clearing the last
+    /// selection. Those aren't "sweep" actions, so advancing would surprise.
+    pub fn toggle_commit_selection_and_advance(&mut self) {
+        let cursor = self.commit_list_cursor;
+        let was_selected = self.is_commit_selected(cursor);
+        self.toggle_commit_selection();
+        let now_selected = self.is_commit_selected(cursor);
+        if was_selected || !now_selected {
+            return;
+        }
+        if let Some((_, end)) = self.commit_selection_range {
+            while self.commit_list_cursor <= end {
+                let before = self.commit_list_cursor;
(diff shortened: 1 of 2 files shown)


agavra (commented):
this is awesome, love the addition. I tried out some of the weird edge cases (moving down to 'show more commits' and moving up, into selections, etc...) all work. 

The behavior is a bit weird when "selecting up", do you know what `jjui` does for that? (e.g. if I am selecting commits above the selecting one because I'm "working my way up"). I think in that situation we should just not move the cursor at all.

jaredramirez:
Here is jjui's behavior:

https://github.com/user-attachments/assets/11c1e85e-b19f-4db8-b446-bfa75ee32be9

It jumps down to commit after the last selected commit. I'm down for your suggestion of not moving the cursor in that case at all though!

Then, also, would you be open to changing the "deselect in the middle of a block" behavior? Currently (both in main & this PR), if you want to deselect in the middle of a block, tuicr shortens the block to the commit you deselected, but does not deselect _that_ commit. For example:

https://github.com/user-attachments/assets/6146c542-050e-49ec-adc5-9f4116af60b8

IMO the extra deselect at the end feels a bit weird. I kept this PR as preserving that behavior, but if you are game to change it, I can include that too (totally fine either way, just throwing it out there)

agavra:
Yup! I'm open to your suggestion about deselecting the inclusive range. Thanks :)

jaredramirez:
Great! Ready for re-review 😄

agavra (approved):
Thanks for the contribution @jaredramirez :)

PR #283, opened 2026-05-12
Add tracing profile instrumentation

iheanyi:
Splits the profiling work out of #281 and switches it to tracing-backed spans instead of the previous custom profiler.

`TUICR_PROFILE=1` writes to the temp profile log, `TUICR_PROFILE=/path/to/log` writes to that path, and `TUICR_PROFILE_FILE` can override the default file path. The instrumentation covers startup, VCS detection, diff loading, revision resolution, recent commits, and commit-selection confirmation without writing to the TUI terminal.

Validation:
- `cargo fmt --all --check`
- `cargo test`
- `cargo clippy --no-default-features -- -D warnings`
- `git diff --check`
- Smoke-tested `TUICR_PROFILE` file output with `target/debug/tuicr -w --stdout`

### DIFF
--- a/src/profile.rs
+++ b/src/profile.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 use std::sync::OnceLock;
 
+use chrono::{DateTime, Utc};
 use tracing::field;
 use tracing_subscriber::fmt::format::FmtSpan;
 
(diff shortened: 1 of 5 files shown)


src/profile.rs
```diff
@@ -0,0 +1,97 @@
...
+    if normalized.is_empty() || matches!(normalized.as_str(), "0" | "false" | "off" | "no") {
+        return None;
+    }
+
+    if let Some(path) = std::env::var_os("TUICR_PROFILE_FILE") {
+        return Some(PathBuf::from(path));
+    }
+
+    if matches!(normalized.as_str(), "1" | "true" | "on" | "yes") {
+        return Some(std::env::temp_dir().join("tuicr-profile.log"));
```
agavra: nit: let's include the start timestamp in the filename (e.g. tuicr-profile.<timestamp>.log) so it doesn't overwrite subsequent ones

agavra (approved):
LGTM! Thanks @iheanyi just one comment about file names

PR #285, opened 2026-05-12
Add sparse Git CLI backend

iheanyi:
Replaces #281's sparse-checkout support with a separate Git CLI backend selected once during Git discovery.

This is part 3 of the #281 replacement stack:
- #283 isolates tracing/profile instrumentation.
- #284 splits the existing libgit2 implementation into `Libgit2Backend`.
- This PR adds `GitCliBackend` and uses it for sparse checkout / sparse-index repos.

What changed:
- Keeps standard Git on the libgit2 path by default.
- Adds `backend = "libgit2" | "cli"` config, with `libgit2` as the normal default and `cli` available as an explicit override.
- Detects sparse Git mode at startup and uses the CLI backend all-or-nothing when the selected backend does not support sparse checkout.
- Adds `VcsBackend::supports_sparse_checkout()` so backend capability checks are centralized instead of inferred at each call site.
- Implements CLI-backed working tree, staged, unstaged, commit range, commits + working tree, revision resolution, commit metadata, context expansion, and staging paths.
- Adds a cheap staged/unstaged status probe so sparse repos do not have to parse/highlight full diffs on startup.
- Narrows untracked scans to simple sparse-checkout cones when possible.
- Surfaces startup warnings for sparse CLI routing and for sparse repos without `core.untrackedCache`.
- Streams `git diff --binary` output into the shared parser instead of buffering the full patch.
- Fixes `GIT binary patch` parsing from the original review feedback.

This addresses the original maintainer feedback:
- https://github.com/agavra/tuicr/pull/281#discussion_r3229513530
- https://github.com/agavra/tuicr/pull/281#discussion_r3229615387

Validation:
- `cargo fmt --all --check`
- `cargo test`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo build`
- `git diff --check`
- Dogfooded `target/debug/tuicr -w` and the `$tuicr` wrapper against a local LLVM sparse-index checkout with a temporary modified file inside the sparse cone; the checkout was restored clean afterward.

Note: the LLVM sparse checkout is local dogfood only and is not cloned or used in CI.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,10 +24,12 @@ src/
 ├── vcs/                 # VCS abstraction layer
 │   ├── mod.rs           # detect_vcs(): auto-detect VCS (jj first, then git, then hg)
 │   ├── traits.rs        # VcsBackend trait, VcsInfo, VcsType, CommitInfo
-│   ├── diff_parser.rs   # Unified diff text parser (shared by hg/jj)
+│   ├── diff_parser.rs   # Unified diff text parser (shared by hg/jj and Git CLI)
 │   │                    # DiffFormat enum: Hg (with timestamps), GitStyle (jj/git patches)
-│   ├── git/             # Git backend (uses native git2 library, not diff_parser)
-│   │   ├── mod.rs       # GitBackend: wraps git2 library
+│   ├── git/             # Git backend selector (libgit2 by default, CLI for sparse/opt-in)
+│   │   ├── mod.rs       # GitBackend wrapper, backend preference, repo mode detection
+│   │   ├── cli.rs       # Git CLI backend used for sparse checkouts and backend = "cli"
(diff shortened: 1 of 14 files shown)


iheanyi:
Hey @agavra, I split the sparse-checkout work from #281 into smaller PRs based on your feedback:

- #283 isolates the tracing/profile instrumentation
- #284 introduces the libgit2 Git backend wrapper/refactor
- #285 adds the sparse-checkout Git CLI backend as an all-or-nothing backend

#285 references the original review threads, including the backend-shape feedback and the `GIT binary patch` parser issue. All three PRs are currently green and have clean merge state. Happy to adjust the split or ordering if you want these stacked differently.

agavra (approved):
Thanks @iheanyi - great to have support for sparse indexes :)

PR #287, opened 2026-05-13
add more Z viewport positioning commands

natebrennand:
Adds support for `zt` (move cursor to top) & `zb` (move cursor to bottom):

https://github.com/user-attachments/assets/95fe26f7-9b46-4d8f-bf4f-a5d612e5a2b0


agavra (approved):
Another VIM binding I didn't know existed (I've only used `zz` frequently). Thanks @natebrennand

natebrennand:
> Another VIM binding I didn't know existed (I've only used `zz` frequently). Thanks @natebrennand

I do think that I've fully replicated the settings that I use now 😁

Grateful for the foundation that you put in place, this has totally changed my workflow.

Issue #125, opened 2026-01-20
Support fetching/pushing reviews to forges (GitHub notably)


YPares:
I thought a bit about how we could use `tuicr` as a general PR review tool, by using Github API to fetch & push reviews, as @agavra it seems you told me that's something you could be interested in supporting.

After investigating, it turns out the github API provides entrypoints to fetch and push reviews (in JSON format). I don't know about other forges but a first GitHub integration (requiring nothing more than adding a dependency on `gh`, the advantage being that we wouldn't have to deal with auth ourselves) could already be pretty useful.

- GitHub Docs: [REST API for Pull Request Reviews](https://docs.github.com/en/rest/pulls/reviews)
- GitHub Docs: [REST API for Pull Request Review Comments](https://docs.github.com/en/rest/pulls/comments)

There's a [quick implem plan](https://github.com/YPares/tuicr/blob/todo-yves/plan-forge-integration.md) in my tuicr's fork.

dkarter:
I came to open an issue just like this one. I tuicr is incredible (!!!) I would love to use it for reviewing other people's code (on their PRs - with the same "conventional comments" style - which saves me a bunch of time and helps frame the comment) as well as pulling comments left by others/bots on my PRs and be able to respond to them / select the ones I want to send to the agent.

This will really streamline code reviews

agavra:
@dkarter thanks for the feedback! It's been a while since this has opened up, and given the recent...reliability and slowness of GitHub I've come around to dedicating more time to support a feature like this. 

I'd love for someone to take a stab at it (cough, cough @YPares 😉), but if not I'll see if I can find some time to have a good remote forge integration.

YPares:
Yeah ^^... These past months I've been thinking about that too, but work got mostly in the way. It's also why I've been making close to no PRs for tuicr lately.

One thing I've been thinking about was whether tuicr was the right place to implement individual forge support, or if an external crate providing an abstraction would be a better place.
[...]

PR #288, opened 2026-05-13
perf: streaming highlight + perceived-latency fixes for large diffs

ptoussai:
Four perf/UX fixes that surfaced together while reviewing a 565-file commit (~47k lines, mostly JS/Vue/TS) in a large private repo. tuicr was slow to load, sluggish when switching files, and scrolling was effectively unresponsive (had to kill the process).

Each commit handles one concern. Per-commit messages have the details; quick recap:

- **`perf(syntax): stream syntax highlighting on a background worker`** — parse emits jobs, a background worker highlights them in parallel and streams spans via mpsc; the main loop patches `diff_files` between frames. File list + first diff structure appear after parse alone instead of parse + full sync highlight. Worker spawns lazily on the first drain so `App::new`'s post-`install_diff` reorderings (commit-message insert, directory sort) settle before jobs are resolved against `diff_files` by `syntax_path`.

- **`perf(syntax): prioritize highlight jobs around the focused file`** — pending jobs live in an `Arc<Mutex<VecDeque>>`. When the user jumps files, the App reorders the remaining queue by file-index distance so visible work runs next instead of grinding files near index 0 that aren't on screen.

- **`perf(input): drain queued events before each redraw`** — the loop used to read one event then redraw; on a heavy diff, holding `j` or wheel-scrolling queued 200-600ms of input lag, worst case on direction reversals. Now drains via `event::poll(Duration::ZERO)` after the blocking poll (64-event cap) then redraws once. Atomic-action keys (Ctrl+C, ZQ, dd, `;X`) keep their per-event `continue 'main` shortcut.

- **`fix(ui): preserve manual file-list scroll when diff panel is focused`** — wheel-scrolling the file list was snapping back because the render-time "sync selection to diff's current file" path re-selected each frame and ratatui's `List` auto-adjusted offset to keep selection visible. Track a `manual_file_list_scroll` flag, set on wheel and cleared when `current_file_idx` changes for any other reason. Pre-existing; visible only once the rest of the loop is fast enough.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2540,7 +2540,9 @@ impl App {
         use std::path::Path;
 
         if idx < self.diff_files.len() {
+            let prev = self.diff_state.current_file_idx;
             self.diff_state.current_file_idx = idx;
+            self.reprioritize_highlights_if_file_changed(prev);
             self.diff_state.cursor_line = self.calculate_file_scroll_offset(idx);
             let max_scroll = self.max_scroll_offset();
             self.diff_state.scroll_offset = self.diff_state.cursor_line.min(max_scroll);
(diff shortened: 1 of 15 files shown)


agavra:
@ptoussai I'm not sure I want to add concurrency to tuicr in order to serve use cases with 47k line PR reviews. I'm highly suspect anyone is reviewing a PR that large by hand anyway, and this makes the potential for bugs significant. 

In my experience, loading even 10-20k line rust change is pretty speedy. Maybe because we don't need the full file highlighter the way `vue` does after the previous change.

I'd rather just default to collapsing large file diffs the way github does at some point.

ptoussai:
The change in question did not have any vue files actually.
I did review this PR and yes, it’s unusual, but I think it’s still reasonable for the tool to not hang on it.
I see your point about the concurrency your comment on my last PR seemed to indicate that you wanted to go that way? Cf https://github.com/agavra/tuicr/pull/280#pullrequestreview-4267671753
This PR does this and other fixes.
There are alternatives ways (like lazy/smart loading) if you think concurrency would introduce too much problems down the line.

ptoussai:
Btw, I’m only using this change as an easy way to test the perf myself, as it’s not easily tested in a headless way. A worst case scenario for sure, but also indicative that smaller changes will have less than ideal performance

agavra:
>  I did review this PR and yes, it’s unusual, but I think it’s still reasonable for the tool to not hang on it.

I agree, which is why I suggested things like collapsing large file diffs by default (like GitHub behavior). 

I'm curious if were you running it in release mode or debug mode? It makes a big difference when you run with `--release`. I haven't had any experience, even with very large PRs, of the tool hanging. Though again, I don't have code that requires the "full file context" so I'm guessing that was part of the perf issue.

> I see your point about the concurrency your comment on my last PR seemed to indicate that you wanted to go that way?

I admit I didn't fully think that through, what got me thinking about that was testing this PR locally and noticing it "flash" as things were being processed. So the initial load looked different than the final load. 

>  There are alternatives ways (like lazy/smart loading) if you think concurrency would introduce too much problems down the line.

I'd rather start with the simplest things just to make sure the tool doesn't crash before we get carried away with fancy perf optimizations. Not sure lazy/smart loading is any better than concurrency. 

I think if we show that there really is a need for faster loading then I'm ok with having a concurrent highlighting job but I'd rather it be blocking on render instead of streaming results back.

ptoussai:
That seems reasonable. The perf issue was not due to debug mode or container (vue etc) files but maybe something else is going on with that change.
The 2 other fixes also really help so I could start with this in place.
I will try to rework it next week.

agavra:
@ptoussai looks like for #125 I'm going to need to add a bunch of async loading stuff anyway, so I'm more open to adding concurrent loading for the main path as well. thanks for your patience with my back and forth here! I'll ping you back when that line of work is done since it's really invasive

PR #295, opened 2026-05-14
forge: existing remote GitHub comments, read-only (PR 4 of forge v1)

agavra:
## Summary

PR 4 of the [forge integration v1 spec](../blob/main/docs/forge-integration-v1-spec.md). Adds read-only display, filtering, and export of existing GitHub review discussions in PR mode.

### In scope
- `RemoteReviewComment` and `RemoteReviewThread` domain types under `src/forge/remote_comments.rs` with `PrCommentsVisibility` (`Unresolved` default, `All`, `Hide`).
- `ForgeBackend::list_review_threads`; GitHub impl uses `gh api graphql` with pagination (`endCursor`/`hasNextPage`, hard-capped at 100 pages).
- Async fetch on PR open and `:e` reload via `spawn_pr_threads_fetch` + `poll_pr_threads_events`. The PR diff renders immediately; threads fade in once the fetch lands. Stale results are discarded by matching against the current PR's repo/number/head SHA.
- `:comments unresolved | all | hide` commands; visibility persists per session.
- Inline rendering in both unified and side-by-side diffs. `[github @alice]`, `[github @bob resolved]`, `[github @bob outdated]` badges; replies indented under the root with `↳`.
- PR-mode markdown export gains an `## Existing GitHub Comments` section grouped by file (unresolved only; resolved/outdated are omitted from export per spec).
- `Loading remote comments…` hint in the status bar while the fetch is in flight.
- Help popup updated with the three new commands.

### Out of scope (explicitly punted)
- Replying / resolving / unresolving threads.
- Editing / deleting remote comments.
- `:submit*`, locked comments, resolver UI (PR 5/6).
- README repositioning (PR 7).
- New CLI args.

### GraphQL vs REST
v1 needs `:comments unresolved` filtering, which requires knowing each thread's `isResolved` and `isOutdated` state. The REST endpoints (\`pulls/comments\` and \`issues/comments\`) do not surface thread resolution state — that lives only on the GraphQL \`PullRequestReviewThread\` node. Per the spec's "GitHub Transport > Review comments and discussions" section we use \`gh api graphql\` with the existing \`GhCommandRunner\` seam so unit tests can drive parsing without touching the network. The same runner abstraction is what PR 1/3 use for everything else \`gh\`-related.

### Annotation parity with rendering
\`AnnotatedLine::RemoteThreadLine\` is pushed in matching count to the rendered comment box so cursor / scroll / search math stays correct. The shared \`thread_display_lines\` helper is used by both the annotation builder and the renderer to keep them in sync — divergence would silently break cursor placement.

### Outdated-line handling
GitHub returns \`line: null\` for outdated comments with the last-known anchor in \`originalLine\`. We fall back to \`originalLine\` so the renderer can still place the thread roughly where it was; the \`is_outdated\` flag drives both the muted styling and the suppression from \`:comments unresolved\`.

### Test seam for integration
\`App::open_pr_with_backend\` now also synchronously fetches threads via the same backend so integration tests can assert on \`forge_review_threads\` without spinning up the background thread + mpsc round-trip. Production paths still use \`spawn_pr_threads_fetch\` for the async behavior.

## Reviewer attention

- Lifecycle math in \`spawn_pr_threads_fetch\` + \`poll_pr_threads_events\` (stale-discard predicate uses repo + number + head SHA — see \`should_discard_stale_remote_threads_event_after_switching_pr\`).
- \`thread_display_lines\` parity with \`comment_panel::format_remote_comment_lines\` — if you change one, change the other or annotations will drift.
- Export filter (\`filter_threads(..., Unresolved)\`) is what enforces "resolved/outdated stays on GitHub". Confirm that matches your reading of the spec's Export Behavior section.
- The \`SyntaxHighlighter\` doesn't cross thread boundaries; we run \`prepare_open_pr\` on the main thread, then spawn the thread fetch separately.

## Test coverage

506 → 541 (+35) tests:

- \`forge::remote_comments\` — visibility filter behavior, threading helpers, serde round-trip.
- \`forge::github::review_threads\` — GraphQL parsing fixtures: single thread, multi-comment thread, resolved, outdated (with \`originalLine\` fallback), cross-file threads, empty array, null repository, malformed JSON, paginated query builder.
- \`forge::github::gh\` — end-to-end \`list_review_threads\` call through \`FakeGhRunner\`, asserts on the \`gh api graphql\` argv shape.
- \`model::review\` — \`remote_comments_visibility\` defaults / round-trips / legacy session JSON compatibility.
- \`app\` — opening a PR populates threads; visibility commands route through the command handler; non-PR mode warning; \`PrThreadsEvent\` apply/discard semantics.
- \`output::markdown\` — PR mode export includes unresolved threads grouped by file; export works with only remote threads; non-PR mode silently omits the section.
- \`ui::diff_unified::remote_comments_snapshot_tests\` — render unresolved comment, render resolved under \`:comments all\` with muted marker, render outdated marker, hide under \`:comments hide\`.
- \`ui::diff_side_by_side::remote_comments_side_by_side_snapshot_tests\` — render badge in side-by-side, hide under \`:comments hide\`.

## Test plan

- [x] cargo fmt
- [x] cargo clippy --all-targets --all-features -- -D warnings
- [x] cargo test (541 passed)
- [ ] Manual: open a real GitHub PR with both unresolved and resolved review threads; verify default view only shows unresolved, \`:comments all\` adds the resolved ones with the muted marker, \`:comments hide\` clears them.
- [ ] Manual: \`:e\` re-fetches threads in place (no diff reparse) on same-head PR.
- [ ] Manual: confirm \`gh api graphql\` works on an enterprise host (\`--hostname\` is set when \`repo.host != "github.com"\`).

### DIFF
--- a/src/forge/github/gh.rs
+++ b/src/forge/github/gh.rs
@@ -203,6 +203,13 @@ where
     }
 
     fn get_pull_request_diff(&self, pr: &PullRequestDetails) -> Result<String> {
+        // We want the *cumulative* diff between base and head for the PR.
+        // `gh pr diff --patch` returns mbox-style `git format-patch` output
+        // — one patch per commit — so a 7-commit PR yields 7 separate
+        // `diff --git` blocks per file, which our parser dutifully turns
+        // into duplicate `DiffFile`s. Plain `gh pr diff` (no `--patch`)
+        // returns the single cumulative diff. Hard-won lesson; see the
+        // duplicate-files-in-list bug.
         self.run_gh(
             vec![
                 "pr".to_string(),
@@ -210,7 +217,6 @@ where
                 pr.number.to_string(),
                 "--repo".to_string(),
                 gh_repo_arg(&pr.repository),
-                "--patch".to_string(),
                 "--color".to_string(),
                 "never".to_string(),
             ],
@@ -872,6 +878,34 @@ index 1111111..2222222 100644
         let patch = backend.get_pull_request_diff(&details).unwrap();
 
         assert_eq!(patch, PR_PATCH);
+    }
+
+    #[test]
+    fn should_not_pass_patch_flag_to_gh_pr_diff() {
+        // Regression: `gh pr diff --patch` returns mbox-style per-commit
+        // patches concatenated, which the diff parser turns into duplicate
+        // DiffFile entries (one per commit-touching-the-same-file). We
+        // want the cumulative diff. Lock the argv so this can't silently
+        // regress.
+        let runner = FakeGhRunner::default();
+        let backend = GitHubGhBackend::with_runner(Some(repo()), runner);
+        let details = backend
+            .get_pull_request(parse_pull_request_target("125").unwrap())
+            .unwrap();
+        let _ = backend.get_pull_request_diff(&details).unwrap();
+
+        let calls = backend.runner.calls.borrow();
+        let diff_call = calls
+            .iter()
+            .find(|args| {
+                args.first().map(String::as_str) == Some("pr")
+                    && args.get(1).map(String::as_str) == Some("diff")
+            })
+            .expect("expected a `gh pr diff` call");
+        assert!(
+            !diff_call.iter().any(|a| a == "--patch"),
+            "`gh pr diff` must NOT pass --patch (mbox output duplicates files); got {diff_call:?}"
+        );
     }
 
     #[test]
--- a/src/forge/github/gh.rs
+++ b/src/forge/github/gh.rs
@@ -2,6 +2,7 @@ use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 
 use crate::error::{Result, TuicrError};
+use crate::forge::remote_comments::RemoteReviewThread;
 use crate::forge::traits::{
     ForgeBackend, ForgeFileLinesRequest, ForgeRepository, PagedPullRequests, PullRequestDetails,
     PullRequestListQuery, PullRequestTarget,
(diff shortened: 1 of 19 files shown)


src/forge/github/gh.rs
```diff
@@ -2,6 +2,7 @@ use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 
 use crate::error::{Result, TuicrError};
+use crate::forge::remote_comments::RemoteReviewThread;
```
agavra, line 5: test comment 1

agavra (commented):
test a top level review comment

src/forge/github/gh.rs
```diff
@@ -220,6 +222,30 @@ where
...
+            all.extend(parsed.threads);
+            let Some(page_info) = parsed.page_info else {
+                break;
+            };
+            if !page_info.has_next_page {
+                break;
+            }
+            let Some(end_cursor) = page_info.end_cursor else {
+                break;
+            };
```
agavra, line 249: test comment with thread

agavra: this is a thread comment

src/forge/github/gh.rs
```diff
@@ -571,10 +628,21 @@ index 1111111..2222222 100644
     impl GhCommandRunner for FakeGhRunner {
         fn run(&self, args: &[String]) -> GhCommandResult<String> {
             self.calls.borrow_mut().push(args.to_vec());
-            match args.get(1).map(String::as_str) {
-                Some("list") => Ok(PR_LIST_JSON.to_string()),
```
agavra, line 575: test comment on old code

src/forge/github/gh.rs
agavra: with a second reply

PR #298, opened 2026-05-14
forge: GitHub review creation + locking (PR 6 of forge v1)

agavra:
## Summary

PR 6 of the [forge integration v1 spec](../blob/main/docs/forge-integration-v1-spec.md). Activates PR 5's local pipeline by wiring the actual `gh api .../reviews` call, plus comment locking and success/error UX. With PR 6 merged, `:submit*` from inside `tuicr pr <target>` posts a real GitHub review.

### In scope
- **`GhCommandRunner::run_with_stdin`** — default trait method (panics if not overridden), `SystemGhRunner` overrides via `std::process::Stdio::piped()`. Lets `FakeGhRunner` capture stdin into a `RefCell<Vec<(args, stdin)>>` for argv+payload assertions.
- **`ForgeBackend::create_review`** + `CreateReviewRequest` + `GhCreateReviewResponse` types. GitHub impl POSTs to `repos/<o>/<r>/pulls/<n>/reviews --method POST --input -`, parses response, maps stderr to spec error messages (missing `gh` / auth / pull-request-write permission).
- **`InlineComment.comment_id: String`** — internal back-reference from a mapped comment to its source `Comment.id`. Selectively excluded from the GitHub JSON payload by the existing builder (regression-tested).
- **Async submit dispatch** — `App::spawn_pr_submit` + `poll_pr_submit_events` + `finish_pr_submit` mirror the `:e` reload pattern. Stale-result discard guards against PR reload mid-submit (repo + PR number + head SHA tuple).
- **Status bar spinner** — `⠋ Submitting review…` / `⠋ Pushing pending review…` while in-flight, right-aligned.
- **Lifecycle writes** — on success, `apply_submit_success` walks `session.files[*]` (file_comments + line_comments) and flips matched comments to `Submitted` (or `PushedDraft` for `:submit draft`), stamps `remote_review_id` from the response. Moved-to-summary items are NOT touched — they embed in the body, not as anchored comments.
- **Session save** — once before spawn (pre-submit snapshot) and again in `finish_pr_submit` after lifecycle writes.
- **Success message** — `"Submitted GitHub review #ID: N inline, M moved to summary"`; draft variant appends `" — Finish it in GitHub: <url>"`.
- **Failure path** — no lifecycle writes, sticky error in status bar, modal stays dismissed, comments remain `LocalDraft`.

### Out of scope (deferred to PR 7 or out of v1)
- Per-comment `remote_comment_id` population (response carries `comments`, but v1 doesn't need them; left `None`).
- Replying to / resolving / unresolving remote threads (out of v1).
- Editing or deleting submitted comments (locked by design; v1 doesn't unlock).
- Pre-submit `gh pr view` refresh (spec hints; deferred — the bg thread does refetch PR details which is good enough).

### `gh api --input -` over CLI args
Multi-comment payloads exceed practical CLI arg length, and stdin lets us serialize the JSON payload directly. The trait extension keeps existing fakes compiling (default panics) without forcing every test to implement an unused stub.

### Stale-result discard tuple
A different PR opened after submit dispatch would otherwise consume the result. The in-flight token snapshots `(repo, pr_number, head_sha)` and we compare on result arrival. Mismatch → silent discard + status-bar info message.

### `comment_id` threading vs parallel arrays
Adding `comment_id: String` to `InlineComment` (vs maintaining a parallel `Vec<String>`) keeps the mapping pipeline simple. The GitHub payload builder uses explicit field serialization so `comment_id` stays internal. A regression test in `github::submit::tests` asserts the field never appears in the JSON.

### Network on bg thread, JSON parse on bg thread too
Unlike PR open (which has to keep the highlighter on the main thread), submit doesn't touch the highlighter. The simpler split is fine: bg thread does network + parse, returns `Result<GhCreateReviewResponse, TuicrError>` over mpsc.

## Reviewer attention

- Stale-result discard predicate in `App::finish_pr_submit` — repo + PR + head SHA comparison.
- Lifecycle write path in `apply_submit_success` — walks both `file_comments` and `line_comments[*]`; verify the comment-id match logic is exhaustive.
- The `parse_create_review_response` keeps `remote_comment_id` as `None` for inline matches. If PR 7 adds "edit submitted comment", we'll need to populate it from `response.comments[*].id` via a path/line match against the request payload.
- Error mapping in `GitHubGhBackend::create_review` — stderr matching is heuristic. The pull-request-write permission case covers both `"Resource not accessible by integration"` and `"must have pull request write"` stderr variants (both have tests).
- Save-before-spawn is synchronous on main thread; if it fails, submit doesn't dispatch and an error is shown.

## Test coverage

613 → 630 (+17) tests:

- `forge::github::gh` — 7 create_review backend tests (success path, draft path, enterprise host, permission/auth/missing-gh failures, malformed response, "must have pull request write" stderr variant)
- `forge::github::submit` — payload regression: `comment_id` never appears in JSON
- `app::submit_flow_tests` — 4 lifecycle (Submitted vs PushedDraft, remote_review_id stamping), 1 only-target-IDs, 2 message-format (published vs draft), 1 failure (sticky error + LocalDraft preserved), 1 stale-result discard, 1 happy-path poll, 1 file-level comment lock

## Hard-won lessons (added to spec)

1. `gh api --input -` is the only practical way to send multi-comment payloads (CLI arg length limits). Test fakes capture stdin alongside argv.
2. Default trait methods keep existing fakes compiling without unused-stub churn — pattern for evolving `GhCommandRunner`.
3. Stale-result discard for submit needs the (repo, PR number, head SHA) tuple, not any single one. A different PR opened mid-submit can otherwise consume the result.
4. `InlineComment → Comment` back-reference is needed for lifecycle writes; the source-comment ID is the only stable anchor (path/line/side could collide).

## Test plan

- [x] cargo fmt
- [x] cargo clippy --all-targets --all-features -- -D warnings
- [x] cargo test (630 passed, 1 ignored)
- [ ] Manual: `tuicr pr <number>`, add a single inline comment, `:submit comment` → review appears on GitHub, comment locks locally (dd shows "Comment already pushed to GitHub").
- [ ] Manual: `:submit draft` → pending review appears on GitHub; success message includes the PR URL.
- [ ] Manual: `:submit approve` and `:submit request-changes` produce the expected GitHub event.
- [ ] Manual: invalid `gh` auth (`gh auth logout` first) → error message matches spec.
- [ ] Manual: simulate a token without write permission → permission error message.
- [ ] Manual: reload mid-submit (open submit modal, `:e` reload before pressing `[y]`) — the in-flight request should be silently discarded if the head changes.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -1,6 +1,7 @@
 use std::ffi::OsStr;
+use std::io::Write;
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum CommandOutputErrorKind {
@@ -43,6 +44,74 @@ where
             status: None,
             stderr: err.to_string(),
         }
+    })?;
+
+    if output.status.success() {
+        Ok(String::from_utf8_lossy(&output.stdout).into_owned())
+    } else {
+        Err(CommandOutputError {
+            kind: CommandOutputErrorKind::Unsuccessful,
+            status: output.status.code(),
+            stderr: String::from_utf8_lossy(&output.stderr).trim().to_string(),
+        })
+    }
+}
+
+/// Variant of `run_command_output` that pipes `stdin` bytes into the spawned
+/// child. Used by `gh api --input -` (and any future tools that want the
+/// same shape).
+pub fn run_command_output_with_stdin<I, S>(
+    program: &str,
+    current_dir: Option<&Path>,
+    args: I,
+    stdin: &str,
+) -> CommandOutputResult<String>
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut command = Command::new(program);
+    if let Some(current_dir) = current_dir {
+        command.current_dir(current_dir);
+    }
+
+    let mut child = command
+        .args(args)
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .map_err(|err| {
+            let kind = if err.kind() == std::io::ErrorKind::NotFound {
+                CommandOutputErrorKind::NotFound
+            } else {
+                CommandOutputErrorKind::SpawnFailed
+            };
+            CommandOutputError {
+                kind,
+                status: None,
+                stderr: err.to_string(),
+            }
+        })?;
+
+    // Write the stdin payload before waiting on stdout, then drop the handle
+    // so the child sees EOF and can finish.
+    if let Some(mut child_stdin) = child.stdin.take() {
+        child_stdin
+            .write_all(stdin.as_bytes())
+            .map_err(|err| CommandOutputError {
+                kind: CommandOutputErrorKind::SpawnFailed,
+                status: None,
+                stderr: err.to_string(),
+            })?;
+        // `drop(child_stdin)` happens when the value goes out of scope, which
+        // closes the pipe and signals EOF.
+    }
+
+    let output = child.wait_with_output().map_err(|err| CommandOutputError {
+        kind: CommandOutputErrorKind::SpawnFailed,
+        status: None,
+        stderr: err.to_string(),
     })?;
 
     if output.status.success() {
--- a/src/process.rs
+++ b/src/process.rs
@@ -52,8 +52,24 @@ where
         Err(CommandOutputError {
             kind: CommandOutputErrorKind::Unsuccessful,
             status: output.status.code(),
-            stderr: String::from_utf8_lossy(&output.stderr).trim().to_string(),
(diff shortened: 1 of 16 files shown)


agavra (commented):
[NOTE] test review leve commit on single commit

<sub>Reviewed with [tuicr](https://github.com/agavra/tuicr).</sub>

src/process.rs
```diff
@@ -51,7 +52,91 @@ where
...
+        })
+    }
+}
+
+/// Build the `stderr` field of a failed-command error from the child's
+/// stdout + stderr. `gh api` puts the JSON response body on stdout even on
+/// non-2xx, while stderr only carries a short status line — surfacing both
+/// (with a separator when both are populated) lets the caller relay the
+/// real API error.
+fn combine_streams_for_error(stdout: &[u8], stderr: &[u8]) -> String {
```
agavra, line 65: [NOTE] test commit on single commit, non-HEAD

PR #303, opened 2026-05-15
forge: resolve SSH host aliases for GitHub remotes

### LINKED ISSUE #302: GitHub PR review fails when git remote uses an SSH host alias
I have two GitHub accounts, personal and work configured as separate SSH aliases in my ~/.ssh/config. When I'm working in a repo that uses my work alias and open the Pull Requests tab in tuicr, it fails with an error because it can't resolve the SSH host alias.

Example:
```
~/.ssh/config:

Host github.com
    HostName github.com
    User git
    IdentityFile ~/.ssh/...

Host github-work
    HostName github.com
    User git
    IdentityFile ~/.ssh/...
```

Image

ikorason:
fix #302 

When a git remote uses an SCP-style URL with an SSH host alias defined in ~/.ssh/config (e.g. `git@github-work:owner/repo.git`), tuicr previously passed the alias to `gh --hostname`, which fails because `gh` only understands real GitHub hostnames. PR listing then errored with "GitHub command failed: error connecting to github-work".

Resolve the alias via ~/.ssh/config when parsing SCP-style URLs, falling back to the alias unchanged when the config is missing, unreadable, or contains no matching block. Only exact `Host` patterns are supported; wildcards, negation, `Match`, and `Include` are documented as out of scope.

### DIFF
--- a/src/forge/github/gh.rs
+++ b/src/forge/github/gh.rs
@@ -1,4 +1,5 @@
 use std::ffi::OsStr;
+use std::fs;
 use std::path::{Path, PathBuf};
 
 use crate::error::{Result, TuicrError};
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thanks @ikorason - very cool to see you using this feature! I haven't fully tested it (I just built this week haha), so I appreciate you working this out 🙏

PR #306, opened 2026-05-15
app: select only cursor commit from all-selected state

### LINKED ISSUE #299: commit selection in diff view is unintuitive
because everything starts selected, it's a bit weird to select something and have it deselect only ranges of things. The first "select" should select only that one and deselect everything else

wnsdud-jy:
## Summary
- When every commit is selected in the inline commit selector, the first toggle now selects only the cursor commit.
- Keeps the existing partial-range shrink behavior for non-all-selected ranges.
- Adds regression coverage for first, middle, and last cursor positions in the all-selected state.

Fixes #299

## Test plan
- [x] `cargo fmt`
- [x] `git diff --check`
- [x] `cargo test commit_selection` — all 4 commit selection tests pass
- [x] `cargo test` — all 402 tests pass, 1 ignored, when run outside the sandbox
- [x] Manual: run `cargo run`, focus the commit selector, press Space on a commit while all commits are selected, and confirm only that commit remains selected

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3816,6 +3816,13 @@ impl App {
                 self.commit_selection_range = Some((cursor, cursor));
             }
             Some((start, end)) => {
+                let all_commits_selected =
+                    self.commit_list.len() > 1 && start == 0 && end == self.commit_list.len() - 1;
+                if all_commits_selected {
+                    self.commit_selection_range = Some((cursor, cursor));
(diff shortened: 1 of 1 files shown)


agavra (approved):
thanks @wnsdud-jy - this had always bothered me as super un-intuitivie. glad the fix was easy 🙏

PR #307, opened 2026-05-15
Support configurable leader key

RyanJamesCaldwell:
## Summary
 
 This PR adds support for configuring tuicr’s leader key in `~/.config/tuicr/config.toml`.
 
 For example:
 
```toml
 leader = ","
```
If no leader is configured, tuicr keeps its existing ; behavior.

## What changed
- Added a new leader config option
- Validated that the configured leader is a single character
- Generalized the existing semicolon-prefixed command handling so it works with 
the configured leader
- Updated in-app help, status bar hints, and docs to reflect the configured 
leader
- Added tests around config parsing and leader-prefixed key handling

## Why

I use a different leader in Neovim and wanted tuicr to feel a little more natural alongside that setup. The existing ;h, ;l, ;c, etc. bindings already behave like leader-prefixed commands, so making that prefix configurable felt like a small quality-of-life improvement.

## A small note

I wrote this with help from Copilot, and I am not especially experienced with Rust. I wanted to contribute anyway, kept the change intentionally focused, and tried to follow the existing structure and conventions in the repo as closely as I could.

I’m very open to feedback on the implementation or on a better Rust-y way to shape any part of it!


agavra (approved):
@RyanJamesCaldwell thanks, this looks like solid rust to me :) most of the tricky rust concepts like ownership/lifteimes don't really apply here

also tried it out locally and it works for me (though I'm not sure why you'd ever want to us `,` as your leader, that's blasphemous! 😆)

PR #318, opened 2026-05-16
fix: pre-wrap comment box content to prevent border overflow

### LINKED ISSUE #205: Multi-line review comment text breaks out of annotation box
Image

When an [ISSUE] annotation comment wraps to a second line, the continuation text is not indented to stay within the annotation box border. In the attached screenshot, the second line ("shutdown.") renders outside/below the left border of the comment box, breaking the visual container.

Expected behavior: Wrapped text in annotation comments should remain indented within the box border.

daveyeb:
## Summary

Fixes #205 — multi-line comment text was breaking out of the annotation box border when ratatui wrapped a long content line; the wrapped continuation had no `│` prefix.

Both the **display** and **input** variants of the comment box now pre-wrap content at the available panel width so ratatui never wraps a line itself. The borders are also dynamic — they always match the panel width instead
of being hardcoded.

The wrap logic is factored into a shared `wrap_segments` helper used by both `format_comment_lines` and `format_comment_input_lines`. Cursor tracking across wrapped visual segments keeps IME screen position correct
when typing past the wrap boundary.

## Test plan

- Open a diff and add a comment with content longer than ~40 chars → comment box border stays intact across multiple visual rows
- Repeat in side-by-side mode (`:diff`)
- Toggle wrap off (`:set wrap!`) — comment box still intact
- Resize terminal narrower while a multi-line comment is visible → border re-flows correctly
- Type a long CJK / Pinyin comment → wraps at correct display width (each CJK char is 2 cols)
- Edit an existing multi-line comment → cursor lands on the correct visual row when navigating

## Out of scope / follow-up

While testing IME input, I noticed that selecting many CJK characters in succession feels slightly slower than typing ASCII. I prototyped a per-line cache that keeps the wrap calc bounded, but the surrounding diff still rebuilds on every keystroke (only made a small difference). Happy to file a separate issue or open a follow-up PR if there's interest.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4297,7 +4297,7 @@ impl App {
     fn review_comments_render_height(&self) -> usize {
         let mut height = 1; // Header line
         for comment in &self.session.review_comments {
-            height += Self::comment_display_lines(comment);
+            height += Self::comment_display_lines(comment, self.diff_state.viewport_width);
         }
         if self.input_mode == InputMode::Comment
             && self.comment_is_review_level
@@ -4324,7 +4324,7 @@ impl App {
 
         if let Some(review) = self.session.files.get(path) {
             for comment in &review.file_comments {
-                comment_lines += Self::comment_display_lines(comment);
+                comment_lines += Self::comment_display_lines(comment, self.diff_state.viewport_width);
             }
         }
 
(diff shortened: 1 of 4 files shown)


agavra (commented):
thanks @daveyeb this almost works but it looks like there are some off by 1 issues. note the letter `h` is missing (and my cursor goes off screen when typing it). 

this is a great UX improvement though, eager to get it in!

daveyeb:
thanks @agavra for the praise. while resolving the off-by-one  issues and doing more QA with CJK content, i realized the annotation count wasn't accounting for visual segments, so wrapped comments fell out of sync with `line_annotations`, breaking `dd` deletion. this started as a small cosmetic fix but ended up touching a bit more than expected.

agavra (approved):

PR #319, opened 2026-05-17
feat: add no_update_check config option

MichaelXavier:
Add a config-file equivalent of the --no-update-check CLI flag so users can persistently skip the startup update check without passing a flag every time.

- Config field: no_update_check: Option<bool> (defaults to None/false)
- CLI flag takes priority via || merge with config value
- Docs updated in CONFIG.md

I dogfooded this by setting the flag in my config then ran `nono learn -- cargo run` and comparing the observed network access against `nono learn -- tuicr`, my global install. The global install did not recognize the flag and so it hit the network to check for updates. The `cargo run` version did recognize the config and did not hit the network.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -25,6 +25,7 @@ wrap = false
 cursor_line = true
 transparent_background = true
 scroll_offset = 5
+no_update_check = false
 
 backend = "libgit2"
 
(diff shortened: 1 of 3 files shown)


agavra (approved):

agavra:
Thanks @MichaelXavier for the contribution!

PR #321, opened 2026-05-17
feat: allow --file to accept a directory for codebase review

davenonymous:
## Summary

Reviewing a whole codebase previously meant relaunching `tuicr` once per file, which is impractical for anything beyond a quick spot-check. The minimal way to unlock multi-file review without inventing a new flag is to relax `--file`: the rest of the app already iterates `Vec<DiffFile>`, so a directory just needs to populate that vector.

Directory mode walks the tree with `ignore::WalkBuilder` so `.gitignore`, `.tuicrignore`, and global git excludes are honored the same way users already expect from the git backend. Binary files are skipped silently and oversized files reuse `is_too_large` with the same 10 MiB threshold as the git backend, keeping the UI's rendering branches uniform across modes.

## Test plan

- `tuicr --file <single-file>` still works as before
- `tuicr --file <directory>` lists all eligible files in the directory tree
- Ignored files (`.gitignore`, `.tuicrignore`, global excludes) are skipped
- Binary files are skipped
- Files over 10 MiB are handled the same way as in git backend mod

## Dog Food
In addition to the integrated tests I've manually ran this against a small-ish and a large repository. It can take a moment to load all files, but that's a problem the other VCS integrations have as well. A large diff touching many files shows the same startup delay. Loading the diff/files asynchronously is out of scope for this PR.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1097,8 +1097,11 @@ impl App {
                 None, // no path_filter
             )?;
 
-            // Hide file list since there's only one file
-            app.show_file_list = false;
+            // Hide the file list only when reviewing a single file; in
+            // directory mode the user needs the list to navigate.
+            if app.diff_files.len() == 1 {
+                app.show_file_list = false;
+            }
             app.focused_panel = FocusedPanel::Diff;
 
             return Ok(app);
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thanks @davenonymous - I confirmed this works for small repos, for large repos (as you mention) it's practically unusable. I agree this isn't unique to this except that it's unlikely you'll have quite as large diffs when reviewing a VCS controlled dir.

That said, we should separately optimize that use case, probably with auto-collapsing files by default

PR #322, opened 2026-05-18
perf(ui): bracket each frame in synchronized output (DEC 2026)

r-vdp:
Wraps each `terminal.draw()` in `BeginSynchronizedUpdate` / `EndSynchronizedUpdate` (`CSI ?2026h` / `l`).

Without this, scrolling a large diff over a slow link (or through a multiplexer like zellij) visibly tears: ratatui's cell-diff escape sequences arrive in chunks and the terminal repaints mid-frame. With the synchronized-update bracket, supporting terminals buffer until the closing `l` and present the frame atomically. Terminals that don't support DEC 2026 ignore both sequences, so this is a no-op there.

crossterm has shipped `BeginSynchronizedUpdate` / `EndSynchronizedUpdate` since 0.26.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,10 +29,10 @@ use crossterm::{
         Event, KeyEventKind, KeyboardEnhancementFlags, PopKeyboardEnhancementFlags,
         PushKeyboardEnhancementFlags,
     },
-    execute,
+    execute, queue,
     terminal::{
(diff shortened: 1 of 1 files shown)


agavra (approved):
nice, thanks again @r-vdp - I wasn't aware of this feature :)

PR #323, opened 2026-05-18
fix(config): honour `wrap = false`

r-vdp:
`DiffState` defaults to `wrap_lines = true`, and the config loader only handled `Some(true)`:

```rust
if cfg.wrap == Some(true) {
    app.set_diff_wrap(true);
}
```

so `wrap = false` in `~/.config/tuicr/config.toml` was a no-op — the only way to start with wrapping off was `:set nowrap` at runtime.

Set `app.diff_state.wrap_lines` directly so both values take effect. This also avoids `set_diff_wrap()`, which prints the "Diff wrapping: on/off" status message — not wanted on startup.


agavra (approved):
thanks for the fix @r-vdp :)

Issue #310, opened 2026-05-15
Add a CLI surface for agent-assisted review workflows


agavra:
## Motivation

`tuicr` users are increasingly reviewing code alongside LLM agents such as Claude and Codex. The current agent flow works, but it still leaves a few rough edges:

- Review feedback often has to move through copy/paste between `tuicr`, the terminal, and the agent.
- Users may want to keep `tuicr` open while an agent inspects the same review state.
- Agents should be able to add draft comments directly into the review, while the human reviewer keeps final control by editing, deleting, or approving those comments before exporting or publishing them in a batch.
- Terminal-first users should be able to keep their existing Claude/Codex + tmux workflow instead of adopting a separate app or frontend.

A small first-party CLI surface would make `tuicr` easier to drive from agents and scripts while keeping the human reviewer in the loop.

## Desired Outcome

Add a CLI-oriented integration path for agent-assisted reviews. At a high level, it should let an external process work with the same review session that the TUI uses: inspect review context, add or update review feedback, mark review progress, and export the final result.

The same underlying capability should also be available from the Rust crate as a library surface. This lets downstream integrations build directly on `tuicr` review behavior without being forced to shell out to the CLI.

The design should preserve the core `tuicr` workflow:

- The TUI remains the primary human review interface.
- The user can review and modify agent-authored comments before they are sent anywhere.
- Review state can be shared between the TUI and CLI so the workflow feels interactive rather than one-shot.
- The integration remains useful to plain terminal users, Claude/Codex users, and other scriptable agent workflows.

## Rejected Alternative: MCP Server In Core

Prior discussion explored MCP/channel support directly inside `tuicr`, but that adds more long-term surface area than this project should take on right now:

- More security and dependency surface, especially if protocol support pulls in new runtimes or larger dependency trees.
- More protocol-specific maintenance as MCP and related client behavior continue to evolve.
- More lifecycle complexity around long-lived server behavior, client compatibility, and bidirectional communication.
- Higher test burden for a project that should stay focused on being a reliable terminal review tool.

A CLI keeps the integration explicit, auditable, and easy to reproduce: commands, arguments, stdout/stderr, exit codes, and session-file mutations can all be inspected directly.

## External MCP Path

This does not rule out MCP support permanently. A separate `tuicr-mcp` project could wrap the CLI and provide MCP-specific behavior for users who want that workflow. If that proves popular and stable, we can revisit whether any of it belongs in the main repository later.

[...]

PR #324, opened 2026-05-18
feat: --all-files for whole-repo annotation

N4M3Z:
Adds `tuicr --all-files` (or `-A`) that lists every git-tracked file in the cwd's repo so one can comment on any line. Same clipboard format as the other modes.

Builds on the directory mode from #321 by @davenonymous. Switches to `git ls-files` so build artefacts don't sneak in, and renders as context instead of additions because nothing is actually changing.

Inspired by revdiff's `--all-files` (https://github.com/umputun/revdiff). Design notes in `docs/decisions/FEAT-0010 All-Files Pristine View.md`.

Notes:
- Mutex with `-r`, `-w`, `--file`. Operates on cwd.
- Non-git invocation errors with a hint.
- Status badges (M/A/D) and side-by-side view are suppressed in this mode — they don't mean anything when nothing's changed.
- Top header chip shows: `PRISTINE · <short-sha> · N files`.
- Session keys are `pristine:<head>:<path_hash>`; loader prefix-matches the hash so a `git pull` mid-review doesn't drop your comments.
- `.tuicrignore` is honored.

Testing:
- `cargo fmt`, clippy clean, `cargo test --bin tuicr` clean
- 15 new tests across `vcs::pristine`, `vcs::file`, `persistence::storage`
- Round-tripped through Claude Code with anchors resolved. `--file .` still renders as additions.

Deferred: `--include`/`--exclude` filters, jj/hg, see the ADR.

### DIFF
--- a/docs/decisions/FEAT-0010 All-Files Pristine View.md
+++ b/docs/decisions/FEAT-0010 All-Files Pristine View.md
@@ -1,0 +1,72 @@
+---
+title: All-Files Pristine View Mode
+description: Add `tuicr --all-files` for whole-repo annotation distinct from `--file` directory mode
+type: adr
+status: proposed
+created: 2026-05-18
+---
+
+# FEAT-0010 All-Files Pristine View Mode
+
+## Context and Problem Statement
+
+tuicr is diff-first: working tree, commit range, GitHub PR. The only non-diff entry point is `--file <path>`, which after #321 also accepts a directory and walks it with `ignore::WalkBuilder`, rendering every line as a new-file addition.
+
+That mode serves "annotate a folder I don't control" workflows, but it doesn't fit whole-repo annotation for a tracked codebase. The walker can include untracked build artifacts when no `.gitignore` is present, and the green `+` gutter on every line implies "review these changes" when the goal is to annotate code that hasn't been touched. Coding agents like Claude Code, Codex, and Cursor want a surface that says "here's the entire codebase, comment on any line, hand the comments back" — a peer to diff modes, not a workaround inside one.
+
+This feature is inspired by [revdiff][REVDIFF]'s `--all-files` mode (umputun's Go-based TUI), which solves the same "annotate every tracked file and feed it to a coding agent" problem in a separate tool. The shape and ergonomics borrow from revdiff; the implementation reuses tuicr's existing review primitives so the new mode composes with vim navigation, comment types, themes, the clipboard export format, and session persistence without changes.
+
+## Decision Drivers
+
+- The clipboard export pipeline (`y` / `:clip` structured markdown) is already what agents consume; the new mode must plug into it unchanged.
+- File enumeration should reflect what is under version control, not a filesystem snapshot.
+- Rendering must signal "no diff here" so reviewers don't mistake context lines for additions.
+- Session persistence must survive a HEAD advance mid-review (e.g. `git pull`) without orphaning comments.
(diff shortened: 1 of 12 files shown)


agavra:
@N4M3Z can you deslop this PR description and explain to me what this is for? I'm not interested in reading verbose AI generated PR descriptions.

N4M3Z:
@agavra Apologies, draft was supposed to stay in my fork until it was ready. Description should be clear now.

docs/decisions/FEAT-0010 All-Files Pristine View.md
```diff
@@ -0,0 +1,72 @@
...
+
+That mode serves "annotate a folder I don't control" workflows, but it doesn't fit whole-repo annotation for a tracked codebase. The walker can include untracked build artifacts when no `.gitignore` is present, and the green `+` gutter on every line implies "review these changes" when the goal is to annotate code that hasn't been touched. Coding agents like Claude Code, Codex, and Cursor want a surface that says "here's the entire codebase, comment on any line, hand the comments back" — a peer to diff modes, not a workaround inside one.
+
+This feature is inspired by [revdiff][REVDIFF]'s `--all-files` mode (umputun's Go-based TUI), which solves the same "annotate every tracked file and feed it to a coding agent" problem in a separate tool. The shape and ergonomics borrow from revdiff; the implementation reuses tuicr's existing review primitives so the new mode composes with vim navigation, comment types, themes, the clipboard export format, and session persistence without changes.
```
agavra, line 17: I'm still not sure I understand the use case (it wouldn't bother me if it didn't add so much code and conditional branching I'd need to maintain going forard).

A human won't want to review that much code in a TUI and agents don't really use TUIs (commenting on all files can be done in a headless way using the upcoming CLI #310) that'll allow them to leverage all the tools they have. 

Am I not understanding something?

agavra (commented):
No worries, happens to me as well. Left a comment, not sure I understand the use case yet.

docs/decisions/FEAT-0010 All-Files Pristine View.md
N4M3Z: Basically, it's about having a look at your entire repo and then adding comments to whichever file you need to. Not just the stuff that's in your commit. Before this I would basically add comments on every tool call and scan the repo as Claude is working. Before that I had to jump around my repo in VS Code and add // TODO comments everywhere or I'd have the remember which files the stuff is in or go tell Claude to find it.

N4M3Z: Maybe this screenshot makes it clearer: `tuicr` -> `c` for comment -> `:clip` -> go, go, go, 
For me a very efficient workflow.

image

(Ofc this would benefit from a single-file view, which I might send your way eventually as well)

agavra (approved):
Thanks for sharing your use case @N4M3Z that makes sense to me now. I tested this locally and when built in `--release` mode it can handle medium sized repos pretty quickly (e.g. `tuicr` itself). Its pretty slow on a larger repo (e.g. https://github.com/slatedb/slatedb shows 400 files, some that are very large) but that's a problem for another time.

thanks for the contribution!

N4M3Z:
Thanks for the quick response, I think naturally a single file view would fix most of the performance problems, I might get to it. I'll send some minor cosmetic things your way in the next couple of days/weeks as I evolve my claude/gitui/tuicr workflow.

PR #327, opened 2026-05-18
add expand after end of diff

natebrennand:
I found myself needing context after the end of the diff frequently.
This PR adds support to expand the file source at the end of the diff.

This wound up being a great deal more involved than anticipated. Included is a subtle bug fix where the expanded chunks on the before/removed diff pane would use line #'s from the added diff side.

Written up via Claude. Lots of manual QA as well as several rounds of Codex reviews.

screenshot w/ expand options:
Screenshot 2026-05-18 at 4 14 06 PM
screenshot w/ expanded diff:
Screenshot 2026-05-18 at 4 14 20 PM

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -51,11 +51,18 @@ fn char_slice(s: &str, lo_char: usize, hi_char: Option<usize>) -> &str {
     &s[lo_byte..hi_byte]
 }
 
-fn gap_annotation_line_count(is_top_of_file: bool, remaining: usize) -> usize {
+fn gap_annotation_line_count(
+    is_top_of_file: bool,
+    is_end_of_file: bool,
+    remaining: usize,
+) -> usize {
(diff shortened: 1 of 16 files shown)


agavra (approved):
tested locally with a few edge cases and looks good to me as well. I did hit a panic underflow for the first time using tuicr but I think that was unrelated to your change. I'll fix that out of band.

There's one test failure, compilation related. Will merge after that.

natebrennand:
> There's one test failure, compilation related. Will merge after that.

Fixing that up, I think my toolchain locally must not be in sync

natebrennand:
> > There's one test failure, compilation related. Will merge after that.
> 
> Fixing that up, I think my toolchain locally must not be in sync

Actually, I just wasn't up to date w/ main. This build should pass.

PR #332, opened 2026-05-19
fix(forge): clear submitted comments from local session on success

agavra:
## Summary

- After a successful `:submit`, every comment that was sent to the forge — inline mappable, summary-bound unmappable, and review-level — is now removed from the local session instead of being marked as `Submitted`/`PushedDraft`.
- This fixes a duplicate-display bug: published inline comments came back via `list_review_threads` on the next `:e` reload, so each just-submitted comment was rendered twice (once as the locked local copy, once as the remote thread).
- `SubmitInFlightState` now snapshots `summary_comment_ids` and `review_comment_ids` at dispatch time so the result handler knows exactly which locals to drop, even if `session.review_comments` grew between preflight and the async result.
- `CommentLifecycleState` / `is_locked` / `remote_review_id` are kept around for backward compatibility with older session files that may already have locked comments serialized.

## Test plan

- [x] `cargo test` (715 passed)
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt`
- [ ] Manual: open a PR, leave inline + review-level + an unmappable-moved-to-summary comment, run `:submit`, then `:e` — verify no duplicates against the freshly-fetched remote threads.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -682,20 +682,31 @@ pub enum PrRangeReloadEvent {
     },
 }
 
-/// Snapshot of the submit state needed to apply lifecycle writes after the
+/// Snapshot of the submit state needed to clear local drafts after the
 /// background `gh api .../reviews` call returns. Captured at spawn time and
 /// stashed on `App::pr_submit_state` so the in-flight spinner has something
-/// to render and the result handler can stamp `remote_review_id` onto the
-/// source comments.
+/// to render and the result handler knows which local comments to drop.
+///
+/// On success every comment that was sent — inline, summary-bound, or review-
+/// level — is removed from the local session, since GitHub now owns them and
+/// a subsequent PR reload would otherwise show duplicates (inline comments
+/// reappearing as remote review threads, plus stale locked drafts cluttering
+/// the diff for items that are now part of the review body).
 #[derive(Debug, Clone)]
 pub struct SubmitInFlightState {
     pub event: crate::forge::submit::SubmitEvent,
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -682,20 +682,31 @@ pub enum PrRangeReloadEvent {
...
-    /// comments, so no lifecycle write applies.
+    /// Source `Comment.id`s of unmappable items the user chose to move into
+    /// the review body's "Unplaced comments" section.
+    pub summary_comment_ids: Vec<String>,
+    /// Source `Comment.id`s of review-level comments that were rendered into
+    /// the review body.
+    pub review_comment_ids: Vec<String>,
+    /// Display count of moved-to-summary items, used only by the success
+    /// message (kept separate from `summary_comment_ids` so message wording
+    /// doesn't accidentally drift if the id list is empty).
```
agavra, line 709: [NOTE] [test comment] why do we need these separate state?

agavra (commented):
<sub>Reviewed with [tuicr](https://github.com/agavra/tuicr).</sub>

src/app.rs
```diff
@@ -5656,6 +5667,14 @@ impl App {
                 }
             })
             .collect();
+        let summary_comment_ids: Vec<String> =
```
agavra, line 5670: [NOTE] test another comment

agavra (commented):
<sub>Reviewed with [tuicr](https://github.com/agavra/tuicr).</sub>

PR #335, opened 2026-05-19
fix split diff gutter sizing in 10k+ line files

natebrennand:
As `app.rs` has grown, it now exceeds 10k lines. This can break down rendering in the git diff if you have changed content with a line number exceeding 4 digits.
This issue did not seem to be present in the unified view, but I did not debug the differences.

before:
Screenshot 2026-05-19 at 1 07 37 PM

after:
Screenshot 2026-05-19 at 1 08 29 PM

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -123,13 +123,40 @@ pub enum ExpandDirection {
     Both,
 }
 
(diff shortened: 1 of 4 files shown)


agavra (approved):
🫠 time to refactor `app.rs` so it isn't >10K lines 😵‍💫

PR #341, opened 2026-05-21
Add support for local TOML themes and custom .tmTheme syntax themes

### LINKED ISSUE #326: Support custom local theme files, including `.tmTheme` for diff syntax highlighting
This feels like a follow-up to #44 rather than a duplicate.

I started looking into adding my preferred theme ([Nightfly](https://github.com/bluz71/vim-nightfly-colors)), and realized the current theme support does not seem to cover the syntax-highlighted code inside the diff.

From reading the existing theme implementation, it looks like the surrounding `tuicr` UI is themeable, but the diff syntax highlighting currently comes from the built-in `bat` / `syntect` theme set. I do not see a way to point `tuicr` at a user-provided syntax theme locally, so even if a built-in UI theme is close, the syntax-highlighted diff still cannot match exactly.

I first looked at upstreaming Nightfly there, but `bat` seems to be discouraging new default themes in favor of local custom themes (for example [sharkdp/bat#3428](https://github.com/sharkdp/bat/pull/3428)), which is understandable from a maintenance point of view. That made me wonder if `tuicr` should support a similar local-theme workflow.

Something like this seems like it could work:
- look for local themes in `~/.config/tuicr/themes/`
- let `theme = "nightfly"` or `--theme nightfly` resolve a local theme as well as a built-in one
- let a local TOML theme file define the `tuicr` UI palette
- let that file optionally reference a Sublime Text `.tmTheme` file for diff syntax highlighting

That seems like it would reduce long-term churn in the built-in theme list while still letting people get an exact local theme match. `tuicr` would not need to parse `.tmTheme` itself either, it could just resolve the path and hand it off to `syntect`.

Let me know what you think!

jackokerman:
## Summary

This PR adds support for loading local TOML themes from tuicr's config directory, including optional local `.tmTheme` files for syntax highlighting.

Theme selection checks bundled themes first, then falls back to `<themes_dir>/<name>.toml`. Local theme files are flat TOML palettes for tuicr UI colors. If `syntax_theme` is set, the referenced `.tmTheme` path is resolved relative to the local theme file.

## Changes

- Added local theme lookup from the config `themes/` directory.
- Added parsing and validation for local theme TOML files.
- Added support for local `.tmTheme` syntax themes, including Base16 ANSI placeholder translation.
- Refactored syntax theme storage to use `SyntaxThemeSource`.
- Updated docs and added a self-contained `tuicr-teal` example theme with a matching syntax theme.

## Notes

Bundled themes remain defined in Rust, and this does not add a new bundled theme. The checked-in `tuicr-teal` files are example local themes that can be copied into the local themes directory.

Closes #326.


examples/nightfly.toml
```diff
@@ -0,0 +1,56 @@
...
+diff_del_bg = "#4e3030"
+diff_context = "#c3ccdc"
+diff_hunk_header = "#82aaff"
+expanded_context_fg = "#4b6479"
+
+syntax_add_bg = "#2a4e57"
+syntax_del_bg = "#4e3030"
+# Optional syntax theme, resolved relative to this file.
+# This Nightfly example expects fly16.tmTheme from bluz71/fly16-bat.
+syntax_theme = "fly16.tmTheme"
```
agavra: if this is the example we should include it in the `examples/` dir as well. FWIW I wasn't able to get that part of the theme working. is there anything special we need for that locally? I could've just been tired last night, so I can try again if it works for you.

agavra (commented):
This LGTM! One comment inline

jackokerman:
> if this is the example we should include it in the `examples/` dir as well. FWIW I wasn't able to get that part of the theme working. is there anything special we need for that locally? I could've just been tired last night, so I can try again if it works for you.

Good call! I updated this to make the example self-contained. I replaced the Nightfly-specific example with a repo-owned `tuicr-teal` example and added the matching `examples/tuicr-teal-syntax.tmTheme`, so the `syntax_theme` reference works out of the box without downloading anything or vendoring a third-party theme. 

Locally, the syntax theme should work as long as the `.tmTheme` file is in the same theme directory as the TOML theme file, or referenced by a relative path from that file.

agavra (approved):
this is great, and I quite like the teal theme! thanks for the contribution @jackokerman

PR #345, opened 2026-05-21
[forge] document forge-specific config settings

natebrennand:
Love the PR review feature addition!

Started trying to add this setting only to learn it already exists. Documenting so others can find it more easily.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -36,6 +36,9 @@ comment_types = [
   { id = "praise", definition = "positive feedback" },
   { id = "nit", label = "nitpick", definition = "small optional tweaks", color = "#d19a66" },
 ]
+
+[forge]
+comment_type_prefix = true
 ```
 
 ## Options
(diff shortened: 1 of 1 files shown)


agavra (approved):
🙏 thanks! note that the GH integration is still pretty early, so there will definitely be some rough edges as you use it. still improving it slowly, but feel free to patch things yourself too :)

natebrennand:
> 🙏 thanks! note that the GH integration is still pretty early, so there will definitely be some rough edges as you use it.

Despite being early, it is still a huge step up for larger PR's.

>  still improving it slowly, but feel free to patch things yourself too :)

Will continue to do so 👍🏻

Am noodling on:
1. the right UX for the submit flow
    - I'd like some way to associate a comment with the review
    - kinda would prefer for it to just work on save/exit rather than a special command, or at least prompt in the same way it asks about the clipboard
2. how to get suggestions working
    - probably similar to comment types?

agavra:
> I'd like some way to associate a comment with the review

you can! `;c` will do a review-level comment 

> kinda would prefer for it to just work on save/exit rather than a special command,

I like it, I think in PR mode exiting should enter the submit flow. makes sense to me

> how to get suggestions working 

ah yes, I think a suggestion path makes sense. I think a visual selection + changing to the suggest comment type would copy the contents of selection into the editor and allow you to make changes. I like it 👍

PR #348, opened 2026-05-21
feat: (optional) arrow-key pane focus and hide gesture

N4M3Z:
Stacked on top of #370 (opt-in folder expansion). Adds slide-focus-between-panes and the optional second-press hide gesture, all gated by the same `arrow_tree_navigation` flag (default off).

> [!NOTE]
> This depends on #370 being merged first. Default behaviour is unchanged for vim users -- `h` / `l` stay as horizontal scroll.

- Diff: `h` at `scroll_x == 0` slides focus to the file list (revealing it if hidden) when `arrow_tree_navigation = true`. Otherwise plain horizontal scroll.
- File list: `l` at `max_scroll_x` on a file slides focus to the diff when `arrow_tree_navigation = true`. Tree-nav behaviours on folders come from #370.
- Sticky-prefix rendering and `scroll_x = 0` reset on leaving the file list are unconditional.
- `right_arrow_hides_file_list = true` enables a press-release-press collapse: the slide arms it, a Right Release sets the gate, a fresh Right Press in the diff collapses the file list. Held keys never satisfy the gate. Requires kitty `REPORT_EVENT_TYPES` to be useful.
- ADR: [docs/decisions/FEAT-0012 Arrow Navigation.md](docs/decisions/FEAT-0012%20Arrow%20Navigation.md)

## Test plan

- [x] Manual: default config keeps `h`/`l` as plain horizontal scroll
- [x] Manual: with `arrow_tree_navigation = true`, `h` in the diff at `scroll_x = 0` slides focus to the file list (and reveals it if hidden)
- [x] Manual: with the same config, `l` on a file in the file list slides focus to the diff
- [x] Manual: sticky prefix (status badge, checkbox, indent) stays visible while filenames scroll horizontally
- [x] Manual: file list `scroll_x` resets to 0 after focus moves away
- [x] Manual: with `right_arrow_hides_file_list = true`, deliberate release+press of `l` in the diff collapses the file list; held `l` does not

### DIFF
--- a/docs/decisions/FEAT-0011 Single-File View.md
+++ b/docs/decisions/FEAT-0011 Single-File View.md
@@ -57,4 +57,4 @@ Entering Comment mode snaps `diff_state.scroll_x = 0` so the inline input box al
 - Pristine `--all-files` over a very large repo loads every tracked file at startup; the continuous scroll is gone but the initial load isn't paginated. A lazy file-tree load (only build the diff for the visible file plus a small window) would scale further.
 - Consider per-session toggle persistence so a user who toggled `:focus` once does not have to re-toggle every launch.
 
-[FEAT-0012]: FEAT-0012 Arrow-Key Pane Focus.md
+[FEAT-0012]: FEAT-0012 Arrow Navigation.md
--- a/docs/decisions/FEAT-0012 Arrow Navigation.md
+++ b/docs/decisions/FEAT-0012 Arrow Navigation.md
@@ -1,0 +1,40 @@
+---
+title: Arrow Navigation
+description: Left/right arrow keys slide focus between the file list and the diff at the scroll boundary
+type: adr
+status: proposed
+created: 2026-05-21
+---
+
+# FEAT-0012 Arrow Navigation
+
+## Context
+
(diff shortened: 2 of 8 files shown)


agavra:
hmm, I don't think I like this to be honest. tried it out locally and I prefer just using the leader for swapping focus `;h` and `;l` is quite simple

N4M3Z:
Yeah, tab works too for that and it might just be sufficient. I was trying to capture what `gitui` does, where on right arrow press the pane "slides" out of the way, you review your file. Leave this open for a bit, I'll try to play with the UX. I might end up just extracting the part where right arrow in the file explorer allows you to quickly auto expand folders and this makes it more easy to navigate to the files you want.

N4M3Z:
Closing in favor of two narrower changes:

- **Sticky-prefix rendering** rolled into #370 (the `arrow_tree_navigation` PR) since it makes the horizontal scroll there look right out of the box.
- **Press-release-press hide gesture** is obsoleted by #373 (Esc hides the file list), which covers the same "I'm done with the chrome" need without depending on kitty `REPORT_EVENT_TYPES`.

The remaining slide-focus-between-panes behavior lives on in my fork at N4M3Z/tuicr#3 for personal use; happy to re-propose upstream as a smaller PR if there's appetite once #370 lands.

PR #350, opened 2026-05-22
Add hunk-level reviewed state

cpaasch-oai:
When reviewing large diffs, marking an entire file as "reviewed" is often too coarse. It is useful to mark only a single "hunk" as reviewed. This PR enables this with the key "R".

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1736,10 +1736,12 @@ impl App {
         path_filter: Option<&str>,
         repo_url_override: Option<ForgeRepository>,
     ) -> Result<Self> {
-        // Ensure all diff files are registered in the session
-        for file in &diff_files {
-            session.add_file(file.display_path().clone(), file.status, file.content_hash);
-        }
+        // Ensure all diff files are registered in the session. Persisted PR
+        // subsets hydrate through the full PR diff first; keep subset-specific
+        // hunk keys alive until the selected diff is loaded.
+        let preserve_hunks = matches!(diff_source, DiffSource::PullRequest(_))
+            && session.commit_selection_range.is_some();
+        Self::register_diff_files(&mut session, &diff_files, preserve_hunks);
 
         let has_more_commit = commit_list.len() >= VISIBLE_COMMIT_COUNT;
         let visible_commit_count = if commit_list.is_empty() {
@@ -2509,14 +2511,37 @@ impl App {
         // Re-register diff files against the loaded session so any new files
         // in the PR appear with content_hash tracking, and any deleted files
         // simply stop appearing in the file list.
-        for file in &opened.diff_files {
-            let path = file.display_path().clone();
-            persisted.add_file(path, file.status, file.content_hash);
-        }
+        // Strict subset sessions are reloaded through a full PR diff first, so
+        // pruning here would discard hunk keys hidden by the active selector.
+        let preserve_hunks = Self::is_strict_commit_selection(
+            persisted.commit_selection_range,
+            opened.commits.len(),
+        );
+        Self::register_diff_files(&mut persisted, &opened.diff_files, preserve_hunks);
         persisted.pr_session_key = Some(key);
         persisted.diff_source = SessionDiffSource::PullRequest;
         persisted.updated_at = chrono::Utc::now();
         opened.session = persisted;
+    }
+
+    fn is_strict_commit_selection(range: Option<(usize, usize)>, total: usize) -> bool {
+        range.is_some_and(|(start, end)| {
+            total > 0 && start <= end && end < total && (start > 0 || end + 1 < total)
+        })
+    }
+
+    fn register_diff_files(
+        session: &mut ReviewSession,
+        diff_files: &[DiffFile],
+        preserve_hunks: bool,
+    ) {
+        for file in diff_files {
+            if preserve_hunks {
+                session.add_diff_file_preserving_hunks(file);
+            } else {
+                session.add_diff_file(file);
+            }
+        }
     }
 
     /// Direct-entry PR open: `tuicr pr <target>`.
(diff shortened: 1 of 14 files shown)


src/model/diff_types.rs
```diff
@@ -73,6 +73,32 @@ pub struct DiffFile {
     pub content_hash: u64,
 }
 
+impl DiffHunk {
+    /// Hunk review state is tied to the exact diff span and content so changed
+    /// hunks ask for review again instead of inheriting stale state.
+    pub fn review_key(&self) -> String {
```
agavra: This is a unique feature, thanks for the suggestion!

The experience seems to be a little rough around the edges so it's probably worth spelling it out the desired behavior. It looks like if anything in a file changes, all hunks after that are marked as unreviewed because the key has changed.

I wonder if the review_key should be a hash of the content of the hunk independent of the start/end line. it's a bit more expensive to compute, but some really cheap hashing function like xxh3 might be worth a try.

I might also be over-designing this. How were you thinking of using this feature?

What do you think?

cpaasch-oai: I like the suggestion of the review_key being a hash of the hunk! It could be quite useful.

I usually use tuicr with `tuicr -r HEAD` then do (small) changes, review those quickly with `git diff` and then merge to HEAD with `git add -u ; git commit --amend --no-edit`. So, recovering the review-state for the larger change does not work as the git-hash changed. Which means, that the review-state recovery was not something I extensively exercised.

So, I basically use "R" to reduce the amount of code I have in front of me (e.g., boilerplate code, includes/imports/...).

I will give your suggestion a shot!

cpaasch-oai: One thing is that different hunks can be the same in different places in a file. So, the hash itself could be problematic. We could increase the context-size for the hunks to make that less likely...

My AI agent also suggested that as a problem. Suggestion is to count occurences of same-hash-hunks and add that to the key. I think that would work...

agavra: 🤔 ah that's a good point. the problem with counting same-hash-hunks is that we can't hash things lazily and I don't really want to be hashing all the hunks up front as I suspect that'll be pretty slow. 

maybe an alternative is using hunk hash + hunk index? that way if a new hunk is added everything gets invalidated, but if you just change within a hunk then only that hunk changes. 

or we can just keep it simple and stick with your original approach. I'm OK either way.

cpaasch-oai: Tried to address this.

agavra: I tested it, I don't think this actually works. here was my test:

1. create three identical hunks
2. mark 1 & 2 as reivewed
3. change 2
4. now hunk 3 is marked as reviewed because it's the second occurrence of the same hash

probably best to just go back to your original approach. sorry for the detour!

cpaasch-oai: Good point! Yes, that doesn't work...

cpaasch-oai: One option is that we could do a hybrid approach, given it is actually pretty rare to have identical hunks:
- If there is only a single hunk with the specific hash in a file, store the hash as key.
- If there are multiple, fall back to line-offset.

How does that sound ? It would make the common case (no identical hunks) work well.

agavra: I like it 👍 I'm a bit worried about the complexity in the storage model for it, but I think we can come up with something clean.

cpaasch-oai: Made the change. LMK what you think!

src/app.rs
```diff
@@ -3962,6 +3975,85 @@ impl App {
...
+            )
+        })
+    }
+
+    pub fn is_hunk_reviewed(&self, file_idx: usize, hunk_idx: usize) -> bool {
+        let Some((path, key)) = self.hunk_review_target(file_idx, hunk_idx) else {
+            return false;
+        };
+        self.session.is_hunk_reviewed(&path, &key)
+    }
```
agavra, line 4166: It looks like the current code hashes `O(Hunk^2)` times per file during the render loop, I feel like that might be a performance consideration so let's just add this quick early-exit that would prevent us from computing the key if this feature is not used at all. We can optimize the path where the feature is used another time.

```suggestion
      // Skip key computation (which hashes every hunk in the file) when this
      // file has no reviewed hunks — the common case for users not using `R`.
      let Some(file) = self.diff_files.get(file_idx) else {
          return false;
      };
      match self.session.files.get(file.display_path()) {
          Some(review) if !review.reviewed_hunks.is_empty() => {}
          _ => return false,
      }

      let Some((path, key)) = self.hunk_review_target(file_idx, hunk_idx) else {
          return false;
      };
      self.session.is_hunk_reviewed(&path, &key)
```

agavra: The full Claude explanation of the H^2 path:

---

**Perf: hunk folding is O(H²·L) per file per render frame, and it fires even when no hunk is reviewed.**

`is_hunk_reviewed` computes the key *before* checking the (usually empty) set:

- `is_hunk_reviewed` (`app.rs:4009`) → `hunk_review_target` (`app.rs:3991`) → `DiffFile::hunk_review_key` (`diff_types.rs:91`)
- `hunk_review_key` calls `hunk_content_hash_counts` (`diff_types.rs:106`), which hashes **every hunk's lines in the file** — O(H·L) — just to resolve one hunk's key.

Render then calls `is_hunk_reviewed` once per hunk (`diff_unified.rs:508`, `diff_side_by_side.rs:541`, annotation rebuild `app.rs:5765`/`9064`), plus up to 2× more per hunk via `should_render_gap_before_hunk` (`app.rs:4016`), plus again in `hunk_positions` (`app.rs:5579`).

→ O(H) calls × O(H·L) each = **O(H²·L) per file, every frame**. Since the hashing happens before the `reviewed_hunks.is_empty()` check, this cost is paid by all users on any diff, whether or not the feature is used.

**Fixes:** (1) early-out in `is_hunk_reviewed` when `reviewed_hunks.is_empty()` for the file — cheap guard that removes the regression for non-users; (2) for the heavy case, hoist the existing batched `hunk_review_keys()` (`diff_types.rs:99`) to once per file/frame instead of calling `hunk_review_key(idx)` in a loop.

cpaasch-oai: Done! Thanks for the suggestion.

agavra:
Thanks for the contribution @cpaasch-oai and for sticking with various rounds of feedback 🙏

cpaasch-oai:
Thanks for merging! I have more PRs coming ;-)

PR #356, opened 2026-05-22
Attribute comment authorship and surface it in the UI

agavra:
## Summary
- Adds an `author` field to local comments (defaults to `"user"`; backward-compatible with existing session JSON).
- Adds `--username <NAME>` to `tuicr review add` (also accepted as `username` in the JSON payload), plus a `username = "..."` config knob that the TUI uses both as the stamp on new comments and as the "viewer" identity for visual distinction.
- In the diff view, non-self comment boxes render with a per-author tinted border and a `[TYPE @author]` badge — mirroring the `[github @author]` style used for remote PR threads. Self comments keep the existing neutral `[TYPE]` style.
- In the comment navigator, non-self entries get a colour-coded `●` chip; the author name is intentionally not shown there to keep rows compact.
- Per-author colour is FNV-1a-hashed onto a 6-slot palette that excludes red/green so it never collides with diff add/del semantics.

## Test plan
- [ ] `cargo test --all-features` (831 passing locally)
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] Open a session in the TUI as `user`; existing comments still render with the neutral border
- [ ] `tuicr review add --session <slug> --username "Claude Opus 4.7" "..."` adds a comment that renders with a tinted border + `[NOTE @Claude Opus 4.7]` badge in the diff view, and a coloured `●` in the navigator
- [ ] Adding a second agent identity (e.g. `--username "GPT-5"`) produces a visibly distinct hash colour
- [ ] Sessions persisted before this change rehydrate with `author: "user"` and keep rendering neutrally

### DIFF
--- a/src/review_cli.rs
+++ b/src/review_cli.rs
@@ -7,8 +7,9 @@ use std::path::{Path, PathBuf};
 use serde::{Deserialize, Serialize};
 
 use crate::cli::{LineSideArg, ReviewCommand};
+use crate::config;
 use crate::error::{Result, TuicrError};
-use crate::model::comment::CommentLifecycleState;
+use crate::model::comment::{self, CommentLifecycleState};
 use crate::model::{Comment, CommentType, LineRange, LineSide, ReviewSession};
 use crate::review_store::{
     AddCommentRequest, CommentTarget, ReviewStore, SessionRef, SessionSummary,
@@ -31,6 +32,7 @@ fn run_with_writer(command: ReviewCommand, out: &mut impl Write) -> Result<()> {
             line,
             end_line,
             side,
+            username,
             content,
         } => add_comment(
             &session,
@@ -42,6 +44,7 @@ fn run_with_writer(command: ReviewCommand, out: &mut impl Write) -> Result<()> {
                 line,
                 end_line,
                 side,
+                username,
                 content,
             },
             out,
@@ -69,6 +72,7 @@ struct AddCommentOptions {
     line: Option<u32>,
     end_line: Option<u32>,
     side: LineSideArg,
+    username: Option<String>,
     content: Option<String>,
 }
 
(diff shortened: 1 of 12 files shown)


src/review_cli.rs
```diff
@@ -7,8 +7,9 @@ use std::path::{Path, PathBuf};
 use serde::{Deserialize, Serialize};
```
agavra, line 7: [NOTE] File-level: File-level note from Reviewer Bot — this should hash to a different palette slot than Claude Opus 4.7, giving you two visually distinct non-self tints to compare side by side.

src/ui/comment_panel.rs
```diff
@@ -310,16 +310,34 @@ pub fn format_remote_thread_lines(
...
+            .fg(styles::author_color_for(name))
+            .add_modifier(ratatui::style::Modifier::BOLD),
+        None => styles::comment_border_style(theme, comment_type.color),
+    };
+
+    let badge_text = match author {
+        Some(name) => format!("[{} @{name}] ", comment_type.label),
+        None => format!("[{}] ", comment_type.label),
+    };
+    let badge_width = badge_text.width();
```
agavra, line 340: [ISSUE] Multi-line range comment to test the [├── …] header path on a non-self author: the badge here will read [ISSUE @Claude Opus 4.7] L336-L340 and the top corner is a tee instead of ╭ since the box anchors to a range.

src/ui/styles.rs
```diff
@@ -113,6 +113,52 @@ pub fn comment_border_style(theme: &Theme, _color: Color) -> Style {
...
+    Color::Blue,
+    Color::LightMagenta,
+    Color::LightCyan,
+];
+
+/// Deterministic palette colour for a given author name. Always returns
+/// a colour; callers gate visibility separately via [`author_accent`].
+/// Hashes via FNV-1a so the mapping is platform-independent and survives
+/// across runs.
+pub fn author_color_for(author: &str) -> Color {
```
agavra, line 133: [SUGGESTION] Consider exposing the palette size so callers can avoid hashing into a slot that collides with diff_add/diff_del fg if a theme remaps them. Probably not necessary today but worth a thought.

agavra (commented):
[PRAISE] Overall this looks tight — the layering between styles::author_color_for (always returns a colour) and styles::author_accent (gates by viewer) is clean and the format_comment_lines change keeps the box semantics consistent with the [github @author] remote-thread style.

[NOTE] Self-authored comment for contrast — this one should render with the neutral border and a plain [NOTE] badge (no @author), so you can A/B it against the agent-tinted boxes above.

PR #362, opened 2026-05-23
fix: use macos-15-intel runner for x86_64-apple-darwin builds

### LINKED ISSUE #361: x86_64-apple-darwin release binary is arm64 (Intel Macs cannot run tuicr)
The `tuicr-0.16.0-x86_64-apple-darwin.tar.gz` release asset contains an `arm64` binary instead of an `x86_64` one, making it impossible to run on Intel Macs.

**Steps to reproduce:**
```bash
brew tap agavra/tap
brew install tuicr
tuicr  # zsh: bad CPU type in executable: tuicr
```

**Verification:**
```bash
$ file $(which tuicr)
/usr/local/bin/tuicr: Mach-O 64-bit executable arm64

$ curl -sL https://github.com/agavra/tuicr/releases/download/v0.16.0/tuicr-0.16.0-x86_64-apple-darwin.tar.gz | tar -xz && file tuicr
tuicr: Mach-O 64-bit executable arm64
```

The `aarch64-apple-darwin` tarball also contains an `arm64` binary (correct), so it appears the same binary was published under both filenames.

**System:** macOS 15.7.4, Intel x86_64 (Kaby Lake)

Johngeorgesample:
## Problem

GitHub switched `macos-latest` to Apple Silicon (ARM) runners in late 2024. Since the build step uses `cargo build --release` without a `--target` flag, it compiles for the runner's native architecture. This means both `x86_64-apple-darwin` and `aarch64-apple-darwin` produced `arm64` binaries, making tuicr unusable on Intel Macs. 

## Fix

Use `macos-15-intel` ([GitHub's recommended Intel x86_64 runner](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/)) for the `x86_64-apple-darwin` build target.

```diff
-   os: macos-latest
+   os: macos-15-intel
```

The `aarch64-apple-darwin` target is unchanged and continues to use `macos-latest` (ARM).

Closes #361

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -154,7 +154,7 @@ jobs:
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
           - target: x86_64-apple-darwin
-            os: macos-latest
+            os: macos-15-intel
           - target: aarch64-apple-darwin
             os: macos-latest
           - target: x86_64-pc-windows-msvc


agavra (approved):
Thanks for the catch @Johngeorgesample :) I don't have an x86 mac to test on so I didn't catch it

PR #364, opened 2026-05-23
feature(provider) add gitlab support

franroa:
Added first attempt to implemet gitlab


src/app.rs
```diff
@@ -723,7 +744,10 @@ pub enum PrRangeReloadEvent {
 }
 
 /// Snapshot of the submit state needed to lock the matching local comments
```
agavra, line 794: codex automated review

[P1] Run rustfmt before merging.

CI runs `cargo fmt --all --check`; on this patch it reports this doc block and other touched files because the new code is not rustfmt-formatted. Until rustfmt is run, the fmt job fails before tests/clippy, and these blank doc-comment lines also trip clippy if left as-is.

src/forge/mod.rs
```diff
@@ -44,3 +46,68 @@ pub fn detect_github_repository(repo_root: &Path) -> Option<ForgeRepository> {
...
+            {
+                all_urls.push(url.to_string());
+            }
+        }
+    }
+
+    // Try GitHub first (filter to hosts containing "github").
+    for url in &all_urls {
+        if let Some(parsed) = parse_github_remote_url(url)
+            && parsed.host.contains("github")
```
agavra: codex automated review

[P2] Preserve GitHub Enterprise remote detection.

With a GitHub Enterprise remote whose host does not literally contain `github` (for example `git@git.company.com:org/repo.git`), `parse_github_remote_url` succeeds but this guard discards it. `App::new_from_pr_target` and the PR selector now use `detect_forge_repository`, so `tuicr pr 123` from those checkouts reports no forge remote even though the GitHub backend supports custom `--hostname`; previous detection accepted these remotes.

src/forge/github/gh.rs
```diff
@@ -606,16 +606,25 @@ fn parse_url_target(target: &str) -> Option<PullRequestTarget> {
     let host = parts.next()?;
     let owner = parts.next()?;
     let repo = parts.next()?;
-    if parts.next()? != "pull" {
+    let segment = parts.next()?;
+    let number = if segment == "pull" {
+        // GitHub: /<owner>/<repo>/pull/<n>
+        parts.next()?.parse::<u64>().ok()?
+    } else if segment == "-" {
```
agavra, line 616: codex automated review

[P2] Let GitLab URLs reach the GitLab parser.

Because `App::new_from_pr_target` tries `parse_pull_request_target` before the GitLab parser, this GitHub parser branch consumes `https://git.company.com/owner/repo/-/merge_requests/42`. For self-hosted GitLab hosts without `gitlab` in the hostname, `forge_repo_from_host` classifies it as GitHub, so the app invokes `gh` instead of `glab`; the GitLab parser never gets a chance.

src/app.rs
```diff
@@ -8797,6 +8837,29 @@ impl App {
...
+            let visibility = self.session.remote_comments_visibility;
+            if !matches!(visibility, PrCommentsVisibility::Hide) {
+                for (thread_idx, thread) in self.forge_review_threads.iter().enumerate() {
+                    if thread.line.is_some() {
+                        continue;
+                    }
+                    let Some(_muted) = visibility.render_decision(thread) else {
+                        continue;
+                    };
+                    let n = thread_display_lines(thread);
```
agavra, line 9013: codex automated review

[P2] Count review-level remote threads in scroll height.

These review-level remote thread rows are now emitted/rendered above the files, but `review_comments_render_height()` and `total_lines()` still only count summaries and local review comments. When a GitLab general MR note is visible, file offsets and `max_cursor_line` are short by these rows, so navigation/scrolling can select the wrong file or prevent reaching the true bottom of the diff.

src/forge/gitlab/glab.rs
```diff
@@ -0,0 +1,1189 @@
...
+            // new_line to resolve the position. For added lines only new_line,
+            // for deleted lines only old_line. counterpart_line carries the
+            // "other side" line number populated by the diff mapper for context lines.
+            let mut position = serde_json::json!({
+                "position_type": "text",
+                "base_sha": pr.base_sha,
+                "start_sha": start_sha,
+                "head_sha": pr.head_sha,
+                "old_path": path,
+                "new_path": path,
```
agavra: codex automated review

[P2] Preserve both paths for renamed GitLab files.

On a renamed GitLab file, `comment.path` is the display/new path from the diff mapper, but GitLab positions require the original `old_path` and the new `new_path`. Setting both fields to the same display path makes inline comments on renamed files invalid or anchored to the wrong file; the mapper needs to carry both paths or the backend needs another way to recover them.

src/forge/gitlab/glab.rs
```diff
@@ -0,0 +1,1189 @@
...
+            });
+            match comment.side {
+                GhSide::Right => {
+                    position["new_line"] = serde_json::Value::Number(comment.line.into());
+                    if let Some(old_line) = comment.counterpart_line {
+                        position["old_line"] = serde_json::Value::Number(old_line.into());
+                    }
+                }
```
agavra, line 453: codex automated review

[P2] Emit GitLab range positions for range comments.

When `comment.start_line` is `Some` for visual/range comments, this payload builder ignores `start_line`/`start_side` and sends only `old_line`/`new_line` for `comment.line`, which is the range end. GitLab will create a single-line discussion instead of the selected range, so range comments on GitLab MRs silently lose their intended anchor.

agavra (commented):
codex automated review

src/forge/github/gh.rs
franroa: This is true. I have no self hosted gitlab though. I think this is not in the scope of this issue. Can we create a separated issue for this?

agavra: Yup, no problem at all for me. I don't have a way to manually test this so I wanted Codex on top of it.

src/app.rs
franroa: done

src/forge/gitlab/glab.rs
franroa: done

franroa:
Hi. Can this be merged?

agavra:
@franroa yup! sorry I was just about to do that then got pulled away for a "real work" thing haha

agavra:
Thanks for the contribution 🙏

PR #365, opened 2026-05-23
perf(ui): skip span allocation for off-screen diff lines

r-vdp:
`render_unified_diff` / `render_side_by_side_diff` currently rebuild every span of every diff line of every file on every frame, then slice the result to the viewport. On a large review (hundreds of files, tens of thousands of diff lines) that's hundreds of thousands of `format!` / `Span::styled` heap allocations per keypress, and scrolling gets visibly laggy.

This gates the per-diff-line span construction on the visible `line_idx` range and pushes a `Line::default()` placeholder for off-screen rows. `lines.len()` therefore still tracks `line_idx` exactly, so the existing `.skip(scroll_offset).take(height)` slice, the cursor/annotation bookkeeping and the comment-box auto-scroll all keep working unchanged. Expanded-context lines get the same treatment.

Per-frame work goes from O(total diff lines) to O(viewport height), which on a 25k-line review is roughly a 500× reduction in allocations.

Two deliberate carve-outs:
- Comment handling after each diff line is left untouched, so a comment box that starts just above the viewport still renders fully into view.
- Comment input mode falls back to the full build because `scroll_comment_input_into_view` may shift `scroll_offset` after the lines are already built.

The first commit is small prep: borrow `line_comments` instead of cloning the per-file `HashMap<u32, Vec<Comment>>` on every frame, and replace the per-header `"═".repeat(40)` with a static `HEADER_RULE`.

### DIFF
--- a/src/ui/diff_side_by_side.rs
+++ b/src/ui/diff_side_by_side.rs
@@ -48,6 +48,16 @@ struct SideBySideContext<'a> {
     // RefCell so deeply-nested rendering helpers can push without each
     // intermediate function needing a `&mut Vec` parameter threaded through.
     comment_bars: std::cell::RefCell<Vec<crate::ui::diff_view::CommentBarAnchor>>,
+    // Only fully build spans for diff lines whose `line_idx` falls in this
+    // half-open range; off-screen rows push `Line::default()` placeholders.
+    visible_start: usize,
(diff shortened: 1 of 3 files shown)


agavra (approved):
awesome stuff, feels better on large diffs.

PR #366, opened 2026-05-23
fix(vcs): detect reftable repos and fall back to CLI backend

saleemdjima:
Closes #236

## Problem
libgit2 refuses to open a repo with `extensions.refStorage = reftable`

## Solution
Detect reftable before attempting libgit2 and route to the CLI backend directly.


agavra:
Thanks @saleemdjima 🙏

PR #370, opened 2026-05-23
feat: (optional) arrow folder expansion in the file tree

N4M3Z:
The default for `h` / `l` is unchanged: they stay as horizontal scroll, matching vim's character navigation. Users who want gitui-style tree nav opt in via config.

- New config `arrow_tree_navigation` (default `false`). When set to `true`, `h` / `l` in the file list fall through to gitui-style tree nav at the horizontal scroll boundary: `l` expands a collapsed folder, descends into an expanded one, or jumps to the next folder below on a file; `h` collapses an expanded folder, otherwise ascends to the parent (and at the top level jumps to the previous folder above).
- With the flag on, moving the file-list cursor (`j` / `k`, arrows, or the tree-nav keys above) to a file also scrolls the diff to that file's header without changing focus. Folders are a no-op so arrowing past collapsed entries leaves the diff alone. Matches gitui / lazygit preview behavior. Enter still commits the selection and shifts focus.
- Horizontal scroll keeps the sticky prefix (indent, expand icon, checkbox, status badge) anchored at the left edge; only the filename portion slides. Scroll caps at "at least one column visible" so a name never disappears entirely. Leaving the file list resets `scroll_x` to 0.
- `Enter` / `Space` still toggle folders directly, regardless of the flag.
- ADR: [docs/decisions/FEAT-0014 Folder Expansion.md](docs/decisions/FEAT-0014%20Folder%20Expansion.md)

## Test plan

- [x] Manual: default config keeps `h`/`l` as plain horizontal scroll
- [x] Manual: with `arrow_tree_navigation = true`, `l` on a folder expands; second `l` descends; `l` on a file jumps to the next folder
- [x] Manual: with `arrow_tree_navigation = true`, `h` on an expanded folder collapses; on a file jumps to the containing folder; at top level jumps to the previous folder
- [x] Manual: with `arrow_tree_navigation = true`, arrowing the file-list cursor across files scrolls the diff to the new file's header; arrowing over folders does not move the diff; focus stays on the file list
- [x] Manual: with `arrow_tree_navigation = false`, arrowing the file-list cursor leaves the diff viewport alone (Enter still required to jump)
- [x] Manual: horizontal scroll on a long filename keeps the checkbox and status badge visible at the left edge; only the name slides
- [x] Manual: switching focus away from the file list and back returns long names to their starting position

### DIFF
--- a/docs/decisions/FEAT-0014 Folder Expansion.md
+++ b/docs/decisions/FEAT-0014 Folder Expansion.md
@@ -1,0 +1,38 @@
+---
+title: Folder Expansion
+description: Opt-in gitui-style folder expand/collapse on h/l in the file list, the diff follows the file-list cursor, and the sticky prefix keeps badges visible during horizontal scroll
+type: adr
+status: proposed
+created: 2026-05-23
+---
+
+# FEAT-0014 Folder Expansion
+
+## Context
+
+Default `h` / `l` are vim's character-navigation keys; in tuicr that maps to horizontal scroll of the focused panel. Some users would rather have those keys drive the file tree the way gitui does -- expand / collapse / descend / ascend. The two intents collide on the same keys, so the gitui behaviour is gated behind a config flag and off by default.
+
+The same audience also expects browse-as-you-go: moving the file-list cursor onto a different file scrolls the diff to that file's header, the way gitui and lazygit preview files. Without it the user has to press Enter to commit each selection, which is awkward when you can already see which file you're on.
+
+Long filenames in deep paths spill past the panel's right edge. The whole row including its status badge slides off under horizontal scroll, leaving the user with no idea which file they're on. The fix is to anchor the badge and indent at the left edge and slide only the name portion.
+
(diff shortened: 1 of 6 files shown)


N4M3Z:
Last larger feature I have in the pipeline, I'll send something cosmetic and that's gonna be it for a while. We should split app.rs into several modules so that we can also split off tests by concern, I might get to it in a couple week

agavra:
Thanks @N4M3Z let's split this into different PRs. I like the mouse down change but I don't want to repurpose `h` and `l` for something other than scrolling, this is primarily meant to "feel" like `vim`, not gitui

also I'll break up `app.rs` soon, that's on my todo list - i totally agree we need that

N4M3Z:
Sure, I'll split it, but please give the folder expansion a shot, it makes it super quick to navigate the folder tree.

N4M3Z:
Minimized the feature set, made the behaviour optional

PR #372, opened 2026-05-24
fix: clipboard copy silently drops content on Linux

ryandielhenn:
On Linux, the clipboard is process-owned. arboard holds ownership via an internal thread, but drops it as soon as the Clipboard object goes out of scope. Since it is dropped immediately after set_text, the content was gone before the user could paste.

This PR adds a fix to try xclip (X11) and wl-copy (Wayland). Whichever is available forks into the background and holds ownership independently of tuicr's lifetime.


agavra (approved):
Thanks! cross platform clipboard support has been pretty challenging

PR #373, opened 2026-05-24
feat: Tab unhides the file list, Esc hides it

N4M3Z:
Split off from the arrow pane navigation PR
- Tab / Shift+Tab unhide the file list when the focus cycle lands on it. A focused-but-hidden panel is invisible state with no way for the user to act on it, so the panel reveals itself whenever focus enters it.
- Esc in any normal-mode panel hides the file list and shifts focus to the diff. `:e` and `<leader>e` still toggle visibility for users who prefer explicit commands.
- ADR: [docs/decisions/FEAT-0016 File List Visibility.md](docs/decisions/FEAT-0016%20File%20List%20Visibility.md)

## Why
- Tab lands focus on the file list when it's hidden, leaving the user on an invisible panel with no cursor; gitui, lazygit, and helix all reveal-on-focus instead of asymmetrically skipping hidden panels.
- Esc already moves focus to the diff in normal mode (not a pure mode-exit), so extending it to also hide the file list adds a one-key dismiss for Tab's reveal without inventing a new keybinding.
- Both behaviors are reversible with `:e` / `<leader>e`, and either can be gated behind a flag if you'd prefer.

## Test plan

- [x] Manual: hide the file list with `:e`, then Tab from the diff -- file list reveals and takes focus
- [x] Manual: hide the file list, then Shift+Tab from the diff backwards -- file list reveals and takes focus
- [x] Manual: with file list visible and focused, Esc hides it and focus moves to the diff
- [x] Manual: with file list visible and the diff focused, Esc hides the file list (focus stays on the diff)
- [x] Manual: `:e` / `<leader>e` continue to toggle visibility regardless of focus

### DIFF
--- a/docs/decisions/FEAT-0016 File List Visibility.md
+++ b/docs/decisions/FEAT-0016 File List Visibility.md
@@ -1,0 +1,28 @@
+---
+title: File List Visibility
+description: Tab reveals the file list when the focus cycle lands on it, and Esc hides it again
+type: adr
+status: proposed
+created: 2026-05-24
+---
+
+# FEAT-0016 File List Visibility
+
+## Context
+
+The file list can be hidden (`:e` toggle, `--no-file-list`, narrow terminal). Once hidden, two awkward states emerge:
+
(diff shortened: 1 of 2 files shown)


agavra (approved):
LGTM! makes sense, thanks for the improvement @N4M3Z

PR #374, opened 2026-05-25
fix: replace deprecated flake output attributes

broady:
Replace deprecated `defaultPackage` and `devShell` flake outputs with `packages.<system>.default` and `devShells.<system>.default`.

These were deprecated in Nix 2.7+ and emit warnings in newer versions. See https://nixos.org/manual/nix/stable/release-notes/rl-2.7


agavra (approved):
Thanks!

PR #376, opened 2026-05-25
feat(vcs): Add whitespace-ignore diffs

abhinav:
Let users opt into whitespace-insensitive local reviews from config:

    ignore_whitespace = true

The `ignore_whitespace` config key now selects all-whitespace
comparison for Git, jj, and hg backends without adding a command
override while command dispatch is being refactored.

PR diffs keep their existing behavior because the forge path does not
yet expose a whitespace-ignore patch contract.

No CLI flags or commands added in this version.


abhinav:

- #376 ◀

agavra (approved):
good stuff, thanks @abhinav for the contribution!

PR #375, opened 2026-05-25
refactor(commands): Add command registry

abhinav:
Command-mode dispatch now resolves fixed commands through a registry
instead of matching every alias directly in the input handler.

This will let some follow-up work take advantage of a global list
of accepted commands, e.g. [`:`-autocomplete](https://github.com/abhinav/tuicr/pull/1) (incoming after this PR):

tuicr-completion

PR #377, opened 2026-05-26
feat: Open focused file in editor

abhinav:
Reviewers can press `e` in normal mode to open the focused
file in `$EDITOR` without leaving tuicr.

The shortcut resolves file-list selections and diff cursor
locations to local worktree files.
Known editors receive best-effort line navigation,
while unknown editors still open the file path.

This required adding one dependency, shlex,
to parse $EDITOR and determine the editor type.
In the future, a configuration could allow specifying
the editor command with a custom template.

---

This PR contains two commits for review clarity:

- First commit refactors terminal session state management
  into a new `TerminalSession` struct.
  This allows easily (and DRYly) adding support for
  suspending terminal session state to shell out to the editor.
- Second commit implements the editor feature.

---

Demo:

tuicr-editor

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -22,6 +22,7 @@ Full reference. Press `?` inside tuicr for an in-app version of this list.
 | `/` | Search within diff |
 | `n` / `N` | Next / previous search match |
 | `Enter` | Expand or collapse hidden context between hunks |
+| `e` | Open focused file in `$EDITOR` |
 | `zt` | Scroll cursor to top of screen |
 | `zz` | Center cursor on screen |
 | `zb` | Scroll cursor to bottom of screen |
@@ -32,6 +33,7 @@ Full reference. Press `?` inside tuicr for an in-app version of this list.
 |-----|--------|
 | `Space` | Toggle expand directory |
 | `Enter` | Expand directory / jump to file in diff |
+| `e` | Open selected file in `$EDITOR` |
 | `o` | Expand all directories |
 | `O` | Collapse all directories |
 
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -22,7 +22,6 @@ Full reference. Press `?` inside tuicr for an in-app version of this list.
 | `/` | Search within diff |
 | `n` / `N` | Next / previous search match |
 | `Enter` | Expand or collapse hidden context between hunks |
-| `e` | Open focused file in `$EDITOR` |
 | `zt` | Scroll cursor to top of screen |
 | `zz` | Center cursor on screen |
 | `zb` | Scroll cursor to bottom of screen |
@@ -33,7 +32,6 @@ Full reference. Press `?` inside tuicr for an in-app version of this list.
 |-----|--------|
 | `Space` | Toggle expand directory |
 | `Enter` | Expand directory / jump to file in diff |
-| `e` | Open selected file in `$EDITOR` |
 | `o` | Expand all directories |
 | `O` | Collapse all directories |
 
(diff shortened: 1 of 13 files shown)


abhinav:

- #377 ◀

docs/KEYBINDINGS.md
```diff
@@ -32,6 +33,7 @@ Full reference. Press `?` inside tuicr for an in-app version of this list.
 |-----|--------|
 | `Space` | Toggle expand directory |
 | `Enter` | Expand directory / jump to file in diff |
+| `e` | Open selected file in `$EDITOR` |
```
agavra: I think we should make this a command `:edit` instead of a keyboard shortcut, since those are pretty limited and reserved for vim-like syntax

src/app.rs
```diff
@@ -3925,6 +3928,93 @@ impl App {
         self.current_file().map(|f| f.display_path())
     }
 
+    /// Takes the queued editor target after action dispatch.
+    ///
+    /// The main event loop consumes this after leaving raw mode and the
```
agavra, line 3989: I think the main UX for the feature should automatically reload the file state (e.g. `:e` which reloads the file) if the worktree is already selected. 

I also think this feature should only be supported if the changes selected are contiguous with the latest changes (e.g. if there are uncommitted changes, the review should include those, otherwise it should include the HEAD commit). Then if it only includes HEAD the reload should automatically reload the session to include the working tree. Otherwise you have this experience where you make a change and can't see it, which is pretty odd.

agavra (commented):
This is a sweet feature @abhinav it's something I've wanted for a while! Added some comments about UX

docs/KEYBINDINGS.md
abhinav: Reasonable. I expected we'd want a command too, but happy to go with command-only.
I was waiting until #375 is (hopefully) merged before trying to add new commands.

agavra: Ah i missed 375, will take a look after my morning meetings

abhinav: No rush on my end. I won't be able to follow up until this evening.

abhinav: This part is done. Personally, I'd still love an `e` keyboard shortcut but I'm happy to defer that to a future version where, say, custom keybindings exist.

src/app.rs
abhinav: > [..] should automatically reload the file state [..] if the worktree is already selected.

Makes sense and straightforward.

> I also think this feature should only be supported if the changes selected are contiguous with the latest changes 

Unless I'm greatly misunderstanding what you're suggesting, I'd like to advise against that:

- "edit this file" (file tree) is desirable regardless of which commit we're looking at (except when it's been deleted)
- "edit at this line" (diff view) _can_ be confusing if your changes aren't visible after or the line numbers are off but it's still convenient for line numbers that are close enough/in the rough hunk range

IMO, it would be limiting to disallow this feature for anything but `-r HEAD` and `-w`,
and very surprising for this to automatically upgrade `-r HEAD` into `-w`

Based on my own experiences:
I reach for edit in contexts of minor changes/amendments I want to make during a review of N commits.
I usually have the files on disk and know how to turn worktree changes into fixup commits.
So line numbers being slightly off from what's in the worktree isn't usually a burden.
When it would be, `git rebase -i` + `edit <hash>` tends to go a long way.
But again—that's just how I've had success using tooling like this.

FWIW, some prior art on this just for comparison:
lazygit relies on a [custom `lazygit-edit://`](https://github.com/jesseduffield/lazygit/blob/7299b224efda4fccea7e62e9c595d2c6b4461639/pkg/gui/gui.go#L378-L379) URI scheme that a user configures their diff pager to use (e.g. `delta --hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}"`).
So delta would use whatever the file/line is in the diff at the time of the diff and generate links for that, regardless of working tree state.

So, I think limiting edit to contiguous changes or expanding diff scope on edit is undesirable.
(But of course it's your call.)

agavra: Fair enough, we can keep the current behavior, just reloading it if the worktree is selected (or adding worktree if it isn't and `..HEAD` is)

abhinav: added reload on worktree selection.
i've kept all commits separate.

agavra (approved):
LGTM! Thanks @abhinav again, you're shipping some really cool features.

PR #379, opened 2026-05-26
feat: (cosmetic) delta-style box separator between files in multi-file view

N4M3Z:
File headers in multi-file view render as a box instead of a single `═══` rule. The cursor skips the entire separator in one keypress.

```
  42 │  }

  ┌─────────────────────────────────────────────┐
  │ ✓ src/app.rs [M]                            │
  └─────────────────────────────────────────────┘
   1 │  use std::path::Path;
```

- New `FileHeaderBorder` annotation for top/bottom box borders. `FileHeader` renders the filename row with closing `│`.
- `cursor_down` / `cursor_up` skip decoration lines (`Spacing`, `FileHeader`, `FileHeaderBorder`) so one keypress crosses the entire separator block.
- `paint_file_header_fill` detects the line type by its leading characters and extends the box to one column before the panel's right edge, with the correct corner glyph (`┐` top, `┘` bottom, `│` middle).
- ADR: [docs/decisions/FEAT-0017 File Separator Box.md](docs/decisions/FEAT-0017%20File%20Separator%20Box.md)

## Test plan

- [x] Manual: multi-file diff shows box separator between files with correct corners
- [x] Manual: box stops one column before the panel border
- [x] Manual: `j` from last line of file N lands on first content line of file N+1, skipping the separator
- [x] Manual: `k` from first content line of file N+1 lands on last line of file N
- [x] Manual: single-file view is unchanged (no box separator)

### DIFF
--- a/docs/decisions/FEAT-0017 File Separator Box.md
+++ b/docs/decisions/FEAT-0017 File Separator Box.md
@@ -14,20 +14,21 @@ The single `═══ filename [M] ═══` rule separating files in multi-fil
 
 ## Decision
 
-File headers render as a three-line box. The box stops one column before the panel's right edge:
+File headers render as a three-line box that spans the panel from edge to edge. The
+indicator gutter is dropped on these rows since the cursor never lands on them:
 
 ```
   42 │  }
 
-  ┌─────────────────────────────────────────────┐
-  │ ✓ src/app.rs [M]                            │
-  └─────────────────────────────────────────────┘
+╔════════════════════════════════════════════════╗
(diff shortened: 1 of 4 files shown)


agavra:
Nice I like this. I've modified the style slightly so it's more obviously not just another comment 

image

we should follow up with this by allowing the cursor to land on that and then `c` would add a file-level comment and file level comments would be displayed inside the box

agavra:
FYI I reverted this change, it takes up 3 lines when 1 will do which makes a big difference in changes with many small changes over many files

PR #384, opened 2026-05-27
fix SBS navigation

natebrennand:
The `zz`, `zb` and expand diffs features were broken on main when using side-by-side diffs. This fixes all the issues I identified and adds some tests that should help mitigate future regressions 🤞🏻.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4505,13 +4505,15 @@ impl App {
     }
 
     pub fn center_cursor(&mut self) {
+        // Use viewport_height (physical rows, stable) rather than visible_line_count
+        // (changes after each render when word wrap is on) so repeated zz presses
+        // are idempotent.
         let viewport = self.diff_state.viewport_height.max(1);
-        let half_viewport = viewport / 2;
         let max_scroll = self.max_scroll_offset();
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -8839,16 +8881,21 @@ impl App {
             }
             let path = file.display_path();
 
-            // File header box (only when shown — same gate as the renderer).
-            // The trailing Spacing from the previous file provides the
-            // blank line above, so no extra Spacing is needed here.
+            // File header: unified mode uses a 3-line box (border+name+border),
+            // side-by-side uses a single ═══ header line.
+            // The annotation count must match the renderer's visual line count.
```
agavra, line 8886: ugh I totally forgot to test this change on the side-by-side mode. I think a better fix here is probably to render the headers the same way (e.g. a box for all of it).

agavra (commented):
LGTM, one thing that's probably worth making consistnetn across SBS and unified instead of accounting for it

src/app.rs
agavra: alternatively I'm happy going tback to just a single header line for unified as well. I don't know if I actually love the new boxes

PR #385, opened 2026-05-27
feat: tab cycle through comments, comment auto-follow, cursor decoration-skip

N4M3Z:
Reworked from scratch on top of current main — the wrap stuff is gone, this no
longer touches diff rendering or line counts.

What's left is the safe half of the original PR:

- Tab cycle goes FileList → Diff → CommitSelector → Comments (reverse mirrors
  it), and tabbing into Comments/FileList unhides the left pane
- j/k in the comment navigator scrolls the diff to the selected comment without
  stealing focus (Enter still jumps + focuses)
- jump_to_file / page scrolls skip spacing + file-header rows so the cursor
  never lands on a non-actionable line — the skip logic is in free functions
  with unit tests
- `:help` / `:h`, and `:wrap` as a short alias for `:set wrap!`

Word-boundary wrap will come back as its own PR, rebuilt against the current
wrap implementation and with rendering tests for the line-count math this time.

Tests: fmt/clippy clean, 972 passing. The 2 failures in `vcs::git::cli::tests`
are pre-existing — they run real `git commit` and inherit the host's
`commit.gpgsign`, so they time out against my YubiKey; same result on a clean
checkout of main.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -457,6 +457,34 @@ pub fn find_source_line(
     }
 }
 
+/// True for rendered lines the cursor should never rest on — spacing between
+/// files and file header rows.
+fn is_decoration(annotation: &AnnotatedLine) -> bool {
+    matches!(
+        annotation,
+        AnnotatedLine::Spacing | AnnotatedLine::FileHeader { .. }
+    )
+}
+
+/// Walk `start` forward (capped at `max_line`) to the nearest non-decoration
+/// annotation so scroll and jump motions land on actionable content.
+fn skip_decoration_forward(annotations: &[AnnotatedLine], start: usize, max_line: usize) -> usize {
+    let mut line = start;
+    while line < max_line && annotations.get(line).is_some_and(is_decoration) {
+        line += 1;
+    }
+    line
+}
+
+/// Walk `start` backward to the nearest non-decoration annotation.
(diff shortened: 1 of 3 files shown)


N4M3Z:
UX fixes, let me know if you want me to split it but most should be sane

agavra:
@N4M3Z I'm going to revert the old line wrapping PR in the meantime until I can battle test this a bit more. it looks like this introduces a few other cosmetic bugs like the file headers being broken 
image

agavra:
I think I need to spend some time figuring out how to have better automated tests because anything that changes rendering line counts causes a lot of downstream problems. see this as well https://github.com/agavra/tuicr/pull/384#discussion_r3307676675

N4M3Z:
Yes, apologies, I'm testing by using so I discover these as I go. I'll contribute here if I find I need to fix more.

N4M3Z:
Reworked this on top of current main — dropped the wrap changes entirely, so
nothing here touches rendering or line counts anymore. Just the panel/cursor
UX bits plus `:help`/`:wrap` aliases.

I'll redo the word wrap as a separate PR against the current implementation,
with rendering tests for the line counts, before asking you to look again.

agavra (approved):
Thanks @N4M3Z this LGTM. I like the comment cycle experience

PR #387, opened 2026-05-27
fix(ui): Preserve wrapped revision headers

abhinav:
Revision views with `wrap = true` render boxed file headers without a
line-number gutter,
but the right-edge fill overlay reused the gutter-adjusted width cache.
In `tuicr -r HEAD`,
that made the fill start before the visible filename ended.

Bad output:

    ║ README                         ║
    ║ docs/KEYBINDING                ║

Expected output:

    ║ README.md [M]                  ║
    ║ docs/KEYBINDINGS.md [M]        ║

Root cause was that the overlay used the wrapped diff-content width:

    ║ README.md [M]
            ^ fill started here and erased `.md [M]`

Fix:
Measure boxed file-header rows from the text they actually render.
This is a bit less performant than the cached gutter-adjusted widths,
but it only applies to the file-header rows.
Everything else still uses the cached gutter-adjusted widths.

### DIFF
--- a/src/ui/diff_unified.rs
+++ b/src/ui/diff_unified.rs
@@ -1346,6 +1346,7 @@ mod remote_comments_snapshot_tests {
     use ratatui::Terminal;
     use ratatui::backend::TestBackend;
     use ratatui::buffer::Buffer;
+    use ratatui::layout::Rect;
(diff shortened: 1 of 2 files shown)


abhinav:

- #387 ◀

PR #394, opened 2026-05-30
perf: streaming highlight + perceived-latency fixes for large diffs (rebase of #288)

r-vdp:
This is #288 by @ptoussai rebased onto current main, since the original branch is still based on pre-v0.13.0 and no longer applies. Authorship is preserved on all four commits.

Supersedes #288.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -945,6 +945,13 @@ pub struct App {
 
     pub file_list_state: FileListState,
     pub comment_navigator_state: CommentNavigatorState,
+    /// True while the user has manually scrolled the file list independently
+    /// of the diff cursor. Suppresses the render-time "sync selection to
+    /// diff's current file" path so wheel scrolls aren't snapped back by
+    /// ratatui's auto-adjust-on-selection-change. Cleared when something
+    /// else changes the diff's current file (cursor moving into a new file,
+    /// click jump, } / { navigation).
+    pub manual_file_list_scroll: bool,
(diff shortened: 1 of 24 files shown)


agavra (commented):
Thanks for rebasing this @r-vdp - I'm curious whether this was something you needed in your local testing. I'm not eager to make the diff rendering a lot more complicated unless people are regularly viewing massive diffs. There are probably a lot more low-hanging fruits to do first before we implement streaming highlighting.

r-vdp:
Yeah, I hit this when I was working on a remote box over ssh on a slow connection, going through diffs was painfully slow.

I have some other, smaller diff, perf improvements sitting locally, I can re-order my patches and submit those instead if you're not eager on merging this one.

agavra:
Yeah let's start with those, I also bet that for slow connections the best thing we can do is auto-collapse some files similarly to how github does it for large diffs since it's mostly about transferring bytes rather than actual loading times (maybe? i'm not actually sure how SSH terminals work in detail). 

I know there's a lot of hype going on recently about Pierre's diffs being blazing fast, but I think that's more novelity than use and I imagine it can create a lot of bugs if we start doing things async.

r-vdp:
The main issue that I encountered was the frame-by-frame rendering being super slow. Every up or down scroll movement would take multiple seconds to rerender the viewport.

I'll review my patch stack and see what else is upstreamable.

PR #395, opened 2026-05-31
fix(git): Preserve revision diff endpoints

abhinav:
`tuicr -r A..B` used to turn the revision range into a list of commits,
then ask the diff loader to reconstruct the file diff from that list.
For linear history this happened to work:
the oldest selected commit's parent is the same tree as `A`.

Merge commits break that assumption.
If `B` merges a side branch,
the oldest selected commit can come from the side branch.
tuicr then compared `parent(side-branch-commit)..B`,
which pulled in changes that already existed in `A`.
The UI showed those left-ref-only changes as reviewable diff content even
though `git diff A B` would not show them.

The bug was that commit selection metadata and diff endpoints were coupled.
The selected commit list is useful for review state and commit metadata,
but it is not a reliable source of the old side of the file diff.

Git revision resolution now preserves both pieces of information:
the selected commits for review metadata,
and the resolved old/new endpoints for diff loading.
Commit selector ranges keep the legacy commit-list target,
while parsed revision expressions carry the user-provided diff boundary.

Testing:
The regression test builds an `A..B` range where `B` is a merge commit.
It verifies that changes already present in `A` stay out of tuicr's diff.

---

As this required a small refactor,
this change is split into two commits:
refactor and fix.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -24,7 +24,7 @@ use crate::vcs::git::calculate_gap;
 use crate::vcs::traits::VcsType;
 use crate::vcs::{
     ChangeKind, CommitInfo, DiffWhitespaceMode, FileBackend, GitBackendPreference, PrNoopVcs,
-    VcsBackend, VcsChangeStatus, VcsInfo, detect_vcs,
+    ResolvedRevisionRange, VcsBackend, VcsChangeStatus, VcsInfo, detect_vcs,
 };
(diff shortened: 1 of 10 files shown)


abhinav:

- #395 ◀

agavra (approved):
Thanks for the fix!

PR #397, opened 2026-06-02
feat: allow :q to quit when only files are marked reviewed

daveyeb:
## Summary

Closes #233.

When a session has **no comments** but files were collapsed as reviewed with `r`, quitting now just works instead of demanding `:q!`:

- `:q` and the `q` keybinding **discard the persisted state and quit** when the only unsaved change is reviewed-file markers.
- Reviewed-only state can be auto-persisted for PR sessions, so the on-disk session file is deleted too — reopening the PR starts clean.
- Sessions with **unsaved comments** are unchanged: `:q` still errors `No write since last change (add ! to override)` and `q` still warns.

## How it works

The decision keys off `session.has_comments()` rather than the coarse `dirty` flag. The only non-comment action that dirties a session is `toggle_reviewed`, so `dirty && !has_comments()` reliably means "reviewed-markers only":

- `dirty && has_comments()` → keep the `:q!` guard / "press q again" warning.
- `dirty && !has_comments()` → discard persisted state and quit.
- not dirty → quit as before.

New `persistence::delete_session_file` removes the deterministically-named session file (missing file = success); `App::discard_session_and_quit` wires it into both quit paths. `ZQ` (explicit force-quit) is left untouched.

## Tests

- `delete_session_file` removes an existing file and treats a missing one as success.
- `:q` quits cleanly when dirty from reviewed-only markers.
- `:q` still blocks when unsaved comments exist.

All 717 tests pass; `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo check` are clean.

## Docs

Updated the `:q` row in `docs/KEYBINDINGS.md`.

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -120,7 +120,7 @@ In command mode,
 | `:clearc` | Clear comments without clearing reviewed marks |
 | `:version` | Show tuicr version |
 | `:update` | Check for updates |
-| `:q` | Quit (warns if unsaved) |
+| `:q` | Quit (warns on unsaved comments; discards review-only state) |
 | `:q!` | Force quit |
 | `:x` / `:wq` | Save and quit (prompts to copy if comments exist) |
 | `ZZ` | Save and quit |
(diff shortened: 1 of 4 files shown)


agavra (approved):

agavra:
Thanks @daveyeb this one has been a bother for a while haha

PR #403, opened 2026-06-03
Fix GitLab MR list pagination flag

ryanmiville:
## Summary
Fix GitLab Pull Requests tab by using `glab mr list --per-page` instead of unsupported `--limit`.

## Test
- `cargo test`


agavra (approved):
Thanks!

PR #408, opened 2026-06-06
Fix crash on repositories with split index enabled

vivekjhaver:
When a repository has `core.splitIndex = true` configured, git uses a
mandatory `link` extension in `.git/index`. Our libgit2 backend does
not support this extension and fails to open the index.

This fix detects the `core.splitIndex` configuration and automatically
falls back to the GitCliBackend, which uses the system git and handles
split indexes correctly.


agavra (approved):
Thanks @vivekjhaver !

PR #412, opened 2026-06-09
fix: map SSH-over-443 transport hosts back to API host

### LINKED ISSUE #411: PR review fails with 'api.ssh.github.com' when ~/.ssh/config rewrites github.com Hostname to ssh.github.com (port 443)
## Summary

`tuicr pr <N>` fails with:

```
Error: GitHub command failed: error connecting to api.ssh.github.com
check your internet connection or https://githubstatus.com
```

when `~/.ssh/config` uses GitHub's documented [SSH-over-HTTPS-port workaround](https://docs.github.com/en/authentication/troubleshooting-ssh/using-ssh-over-the-https-port), which rewrites `github.com`'s `Hostname` to `ssh.github.com`.

`api.ssh.github.com` is not a real host (NXDOMAIN) — only `api.github.com` exists. It looks like tuicr resolves the SSH `Hostname` from ssh config and then prepends `api.` to build the API host, which is incorrect when the hostname has been rewritten for transport reasons.

## Reproduction

`~/.ssh/config`:

```
Host github.com
    Hostname ssh.github.com
    Port 443
    User git
    IdentityFile ~/.ssh/id_ed25519
```

Git remote (vanilla, no alias):

```
origin  git@github.com:Owner/Repo.git
```

Run:

```
tuicr pr <N>
```

→ fails as above.

Workaround that confirms the diagnosis — forcing the repo URL bypasses the SSH-config-derived hostname and works:

```
tuicr pr <N> --repo-url https://github.com/Owner/Repo.git
```

## Environment

- tuicr 0.17.1 (Homebrew)
- macOS (Darwin 25.5.0)
- `gh auth status` healthy, token has `repo` scope

## Related

#302 / #303 fixed the case where `Host` is a *custom alias* (e.g. `github-work`). This is a related but distinct case: `Host` is the real `github.com`, but its `Hostname` is rewritten to `ssh.github.com` for port-443 transport. The fix from #303 likely doesn't cover this because the host name isn't an alias — it's a transport rewrite of a real host. tuicr should either (a) not derive the API host from ssh `Hostname` at all and always use `api.github.com` for github.com remotes, or (b) recognise `ssh.github.com` as an SSH-only transport host and map it back to `github.com` for API purposes.

The SSH-over-443 config is the [officially documented GitHub workaround](https://docs.github.com/en/authentication/troubleshooting-ssh/using-ssh-over-the-https-port) for networks that block port 22, so it's likely to affect more users.

xylophonehero:
## Summary

Fixes #411.

`tuicr pr` failed with `error connecting to api.ssh.github.com` when `~/.ssh/config` used GitHub's [documented SSH-over-HTTPS-port workaround](https://docs.github.com/en/authentication/troubleshooting-ssh/using-ssh-over-the-https-port) (`Hostname ssh.github.com`, `Port 443`). The resolved SSH hostname is correct for git transport but `api.ssh.github.com` doesn't exist — only `api.github.com` does.

After resolving `~/.ssh/config`, normalize known SSH-transport hosts back to their canonical API host:
- `ssh.github.com` → `github.com`
- `altssh.gitlab.com` → `gitlab.com` (GitLab's equivalent)

This composes with the existing alias-resolution from #303 rather than replacing it: legitimate `HostName` overrides (e.g. GHES) still flow through unchanged.

## Test plan

- [x] Unit tests for `normalize_ssh_transport_host` (github + gitlab)
- [x] End-to-end test parsing an SCP-style URL with the exact `~/.ssh/config` from the issue repro
- [x] Existing `forge::` test suite passes (154 tests)
- [x] Manual: run `tuicr pr <n>` with the SSH-over-443 config in place and confirm the API call succeeds

### DIFF
--- a/src/forge/github/gh.rs
+++ b/src/forge/github/gh.rs
@@ -678,6 +678,11 @@ fn forge_repo_from_host(host: &str, owner: &str, repo: &str) -> ForgeRepository 
 /// a `Host`/`HostName` pair that depends on any of those falls back to
 /// the alias unchanged.
 fn resolve_ssh_hostname(alias: &str) -> String {
+    let resolved = read_ssh_hostname(alias);
+    normalize_ssh_transport_host(&resolved).to_string()
+}
+
+fn read_ssh_hostname(alias: &str) -> String {
     let Ok(home) = std::env::var("HOME") else {
         return alias.to_string();
     };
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thanks @xylophonehero! This might be a little brittle but I don't know a better way so it LGTM

PR #414, opened 2026-06-09
review: attribute commit-message comments to their commit

thiblahute:
When reviewing a range of commits one at a time, comments on each commit's message were rendered indistinguishably as `Commit Message:N`, losing track of which commit a comment applied to.

The session indexes comments by path, but the synthetic commit-message file always used the literal path "Commit Message", so comments on different commits' messages collided into a single key. Embed the commit's short id in the path (`Commit Message (<sha>)`) so each commit gets a distinct key and the exported review records which commit every comment belongs to. The TUI headers and file list now surface the same identity.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3397,6 +3397,11 @@ impl App {
     }
 
     /// If we are viewing a single commit, insert a "Commit Message" DiffFile at index 0.
+    ///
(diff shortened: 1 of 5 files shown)


agavra (approved):
nice! thanks for the contribution @thiblahute

PR #417, opened 2026-06-10
fix(markdown): Indent multiline comments

abhinav:
Markdown export should keep multi-line review comments
inside the same ordered-list item.
Before this change, continuation lines were emitted at column zero,
so Markdown parsers treated them as text outside the numbered item.

The exporter now indents continuation lines
according to the active list marker width.

Regression tests cover both single-digit and double-digit list markers
so `1.` and `10.` preserve valid Markdown alignment.


abhinav:

- #417 ◀

agavra (approved):

PR #419, opened 2026-06-13
docs: document GitLab merge request support

bendrucker:
This PR adds documentation for tuicr's GitLab support. The feature has improved steadily (#364, #412, #416) but had no docs.

I'm planning a handful of feature contributions too, but figured I'd start here. I'm primarily a GitHub user, though I've had jobs using GitLab, so I'm motivated to help work toward parity. I also have a personal account I use for testing automation, so I can help with testing and bug fixing.

## What's here

- New `docs/GITLAB.md`: setup with `glab`, opening an MR, submitting Comment and Approve reviews, self-hosted instances, SSH over port 443, and troubleshooting.
- `README.md`: GitLab is now a first-class forge. That covers the tagline, the "what it does" bullets, the comparison table, a new "To GitLab" export subsection, and a quick-start example.
- `docs/CONFIG.md`: the `[forge]` section now covers GitHub and GitLab.
- `docs/KEYBINDINGS.md`: a note on which submit events apply to which forge.

No code changes. This is documentation only.

## Notes

GitLab does have a request-changes review state, but tuicr doesn't currently submit it (or draft reviews) to a GitLab MR. It returns an "unsupported operation" error instead. I documented the behavior as it stands today rather than change the code here. The current error string is also worded as if GitLab lacks the feature, which isn't accurate. I left both alone on purpose and plan to come back for the real parity work in a follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # tuicr
 
-**A code review TUI with vim keybindings. Export to GitHub or clipboard.**
+**A code review TUI with vim keybindings. Export to GitHub, GitLab, or clipboard.**
 
 [![Crates.io](https://img.shields.io/crates/v/tuicr)](https://crates.io/crates/tuicr)
 [![License](https://img.shields.io/crates/l/tuicr)](./LICENSE)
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thanks @bendrucker :) looking forward to your other contributions

PR #421, opened 2026-06-17
feat(gitlab): support request-changes reviews on merge requests

bendrucker:
Adds `:submit request-changes` for GitLab merge requests. GitLab has a "Request changes" reviewer state with no REST endpoint, but it is settable through the `mergeRequestRequestChanges` GraphQL mutation. tuicr now posts the review comments and then runs that mutation to set the MR reviewer state to changes requested, matching the GitHub behavior. The mutation takes the un-encoded project path and the MR iid as a string.

GraphQL returns HTTP 200 even on logical failure, so a successful exit status is not enough. A new `check_graphql_errors` helper parses the response body and surfaces any errors. Only an assigned reviewer can set this state.

Follows up on #419, which flagged this gap. Draft reviews stay GitHub-only and will be my next target.

## Notes

- Comments post before the mutation runs, so a failed request-changes still leaves the inline comments on the MR. The operation is not atomic.
- GitLab records a "requested changes" system note even when the mutation returns an error. Each failed non-reviewer attempt adds one, but the reviewer state never actually changes (`reviewers` stays empty). This is a GitLab-side quirk, not something tuicr controls.
- The error surfaces GitLab's verbatim text (e.g. `Reviewer not found`), which is terse. A friendlier wrapper is possible but out of scope here.
- Out of scope, for later: draft reviews via `draft_notes`, auto-adding the user as a reviewer, and clearing the state (`mergeRequestDestroyRequestedChanges`).

## Testing

New unit tests cover the mutation call sequence (body note, then inline discussion, then the GraphQL mutation carrying the un-encoded `projectPath` and string `iid`), the error surfaced when the mutation returns a non-empty `errors` array, continued rejection of draft reviews, the approve path, and `check_graphql_errors` parsing for payload-level `[String!]` errors, top-level GraphQL error objects, and non-JSON output.

Verified live against gitlab.com through the TUI, using a throwaway project with two MRs:

- Happy path ([MR !2](https://gitlab.com/bendrucker1/tuicr-request-changes-test/-/merge_requests/2)): submitting as an assigned reviewer posted the inline comment and flipped the reviewer state from `UNREVIEWED` to `REQUESTED_CHANGES`.
- Error path ([MR !1](https://gitlab.com/bendrucker1/tuicr-request-changes-test/-/merge_requests/1)): submitting as the author (not a reviewer) posted the comment, then surfaced `Submit failed: GitLab request-changes failed: Reviewer not found` instead of reporting success.

Sandbox project (archived): https://gitlab.com/bendrucker1/tuicr-request-changes-test

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,10 +106,10 @@ Requires `gh` authenticated to the repo.
 
 ### To GitLab
 
-`:submit` offers Comment or Approve on a GitLab MR. Inline comments post as discussion notes.
-Review-level comments become the summary. Requires `glab` authenticated to the host. tuicr submits
-only Comment and Approve to GitLab. Request changes and Draft are GitHub-only here. See
-[docs/GITLAB.md](docs/GITLAB.md) for setup, self-hosted instances, and troubleshooting.
+`:submit` offers Comment, Approve, or Request changes on a GitLab MR. Inline comments post as
+discussion notes. Review-level comments become the summary. Requires `glab` authenticated to the
+host. Request changes needs your account to be an assigned reviewer. Only Draft is GitHub-only
+here. See [docs/GITLAB.md](docs/GITLAB.md) for setup, self-hosted instances, and troubleshooting.
 
 ### To your coding agent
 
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thanks @bendrucker :)

PR #422, opened 2026-06-18
feat(gitlab): support draft reviews on merge requests

bendrucker:
Adds `:submit draft` for GitLab merge requests. The command creates GitLab draft notes (the pending-review primitive) for the body and each inline comment, then stops. The author finishes the review from the GitLab "Submit review" UI. This mirrors a GitHub pending review: the comments live on the forge as editable drafts until the author publishes them.

The draft path reuses the existing position builder. It swaps two things: the endpoint (`draft_notes` for `notes`/`discussions`) and the JSON field (`note` for `body`). Everything downstream of the backend already handled the draft event generically, so the lifecycle transition to pushed-draft and the success message needed no backend-specific changes.

Follows up on #421, which named this as the next target.

## Notes

- Creating drafts is the whole feature. tuicr does not call `draft_notes/bulk_publish`. Leaving the notes unpublished is what makes this a pending review rather than a normal comment submission.
- Draft notes return a numeric `id`, where discussions return a string one. The response extractor now accepts either, so the success message reports a real id instead of `#0`.
- The submit confirmation dialog and success message are now forge-aware through a small `App::forge_display_name` helper. A GitLab MR previously showed "Push pending GitHub review?"; this path only became reachable for GitLab once drafts worked here.
- Out of scope, for later: publishing drafts from tuicr (`bulk_publish`), auto-adding the user as a reviewer, and clearing request-changes state.

## Testing

New unit tests cover the draft path: body and inline notes posting to `draft_notes` with the `note` field and the expected position, a `PENDING` response state carrying the parsed numeric id, no call touching `bulk_publish` or `/publish`, the default Comment path still hitting `notes`/`discussions`, and the GitLab label in the confirmation dialog.

Verified live against gitlab.com through the TUI, using a throwaway MR ([MR !1](https://gitlab.com/bendrucker1/tuicr-draft-notes-test/-/merge_requests/1)):

- Added a summary comment and an inline comment, then `:submit draft`. The confirmation dialog read "Push pending GitLab review?" and tuicr reported success with the MR URL, flipping both comments to pushed-draft state.
- Confirmed on the API that two draft notes existed (body plus inline at `new_line` 5) and zero notes were published.
- Published the drafts from the API to check the position resolves: the inline note landed on line 5.

Sandbox project (archived): https://gitlab.com/bendrucker1/tuicr-draft-notes-test

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -7387,6 +7387,18 @@ impl App {
         self.finish_pr_submit(in_flight, result);
     }
 
+    /// Human-readable name of the forge backing the current PR/MR review.
+    /// Used to keep submit messaging accurate across GitHub and GitLab.
+    pub fn forge_display_name(&self) -> &'static str {
+        match &self.diff_source {
+            DiffSource::PullRequest(pr) => match pr.key.repository.kind {
+                crate::forge::traits::ForgeKind::GitHub => "GitHub",
+                crate::forge::traits::ForgeKind::GitLab => "GitLab",
+            },
+            _ => "forge",
+        }
+    }
+
     /// Apply the create-review result on the main thread. On success: flip
     /// each included `Comment` to `Submitted` (or `PushedDraft` for the
     /// draft event), stamp `remote_review_id`, save the session again, and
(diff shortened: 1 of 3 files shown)


agavra (approved):
Thanks again :)

PR #424, opened 2026-06-20
feat: default to cursor commit when no range selected

joaomendoncaa:
it's extremely frustrating to have to select a commit

if someone presses Enter/Return, I presume there's no possible intent other than choosing the commit already under the cursor in the first place


agavra (approved):
Makes a lot of sense! Can't believe I hadn't thought of it haha. Thanks @joaomendoncaa

PR #426, opened 2026-06-21
fix(forge): show expand row after last hunk in PR review

### LINKED ISSUE #423: Expand down doesn't show after last hunk
We don't show the ability to expand more lines down after the last hunk.

Image

saleemdjima:
This PR fixes #423

  ## Summary

  - Fix the missing expand row after the last hunk in PR review mode
  - Added `file_line_count` to `ForgeBackend` so the forge API is used directly to count lines
  - Fixed `slice_to_diff_lines` in both forge backends to break early instead of iterating up


agavra (approved):
LGTM and tried it out locally. Thanks @saleemdjima :)

PR #428, opened 2026-06-22
perf(main): only redraw when state changed

r-vdp:
Following up on the discussion in #394, this is one of the smaller perf
patches I mentioned, with no async/streaming involved.

The main loop currently calls `terminal.draw()` every 100ms regardless
of whether anything changed. The diff renderer rebuilds every visible
line on each draw, so on large diffs this burns CPU while idle and
ships needless escape sequences over slow links.

This adds a `needs_redraw` flag set by:
- any input event (set before `event::read()` so the many `continue`
  paths in the key handler still trigger a redraw)
- the background update check landing
- message / Ctrl-C-prompt expiry
- persisted-session reload merging external comments
- in-flight PR work (forces a redraw each tick so spinners keep
  animating and results paint as soon as they arrive)

When none of those fire, `terminal.draw()` is skipped entirely.

`clear_expired_message()` and `poll_persisted_session_changes()` now
return `bool` to report whether they touched visible state. The new
`has_pending_pr_work()` helper just ORs the six PR rx slots.

Tested locally on a ~6k-line diff over a slow ssh hop: idle CPU drops
to zero and ratatui no longer re-emits the cursor-line attrs every
100ms.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2063,25 +2063,45 @@ impl App {
         }
     }
 
-    pub fn poll_persisted_session_changes(&mut self) {
+    /// Returns `true` if visible state changed (external comments merged or a
+    /// warning was raised) so the main loop can schedule a redraw without an
+    /// input event.
+    pub fn poll_persisted_session_changes(&mut self) -> bool {
         let Some(interval) = self.review_watch_interval else {
-            return;
+            return false;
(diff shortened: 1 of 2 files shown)


agavra (approved):
thanks!

PR #430, opened 2026-06-24
feat(comment): opt-in vim modal editing in the review comment box

thiblahute:
## What

Adds **opt-in vim-style modal editing** to the review comment text box, backed by the [`edtui`](https://github.com/preiter93/edtui) crate (used as the editing *model + input engine* only — the existing comment-panel rendering is reused).

- New `comment_vim` config key (default `false`).
- `:vim` command (plus `:set vim` / `:set novim`) to toggle at runtime.
- Header shows the current mode (`[NORMAL]`/`[INSERT]`/`[VISUAL]`).

When off, the box keeps its current emacs/readline bindings unchanged — zero behavior change for existing users.

## Design

`comment_buffer` / `comment_cursor` stay **canonical** (rendering and all read-sites consume them). When vim is on, an `edtui` overlay is built lazily on the first key in comment mode and synced back into the buffer/cursor after each event, with UTF-8-aware `Index2` ↔ byte-offset conversion. This keeps the inline-in-diff preview and IME cursor positioning working with no renderer changes.

App-level keys keep their semantics: `Ctrl-S`/`Ctrl-Enter` save, `Tab`/`Shift-Tab` cycle the comment type, `Ctrl-C` cancels; from Normal mode `Esc` cancels and `Enter` saves. Everything else drives the modal editor (`hjkl`, `w`/`b`/`e`, `dd`/`D`/`ciw`/`x`, `u`/`Ctrl-r`, visual `v`+`y`/`d`/`p`).

## Known limitation

`edtui`'s vim mode binds a fixed set of key *sequences* rather than a full operator+motion grammar, so combos like `dw`/`de`/`cw`/`d$` are not supported yet (use `dd`/`D`/`ciw`/`x`). Upstream [edtui#63](https://github.com/preiter93/edtui/pull/63) adds `dw`/`dW`/`diw`; once released, a dependency bump picks it up. We chose not to work around it locally.

## Testing

- `cargo build`, `cargo clippy --all-targets` (0 warnings), `cargo test` (full suite green).
- New unit tests in `src/comment_vim.rs`: UTF-8 `Index2`↔byte round-trips (CJK/emoji) and integration tests driving real key events through the editor (insert typing, `Esc`→Normal, `x`, `dd`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,10 +635,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "75b325c5dbd37f80359721ad39aca5a29fb04c89279657cffdda8736d0c0b9d2"
 
 [[package]]
+name = "edtui"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4941adefb4c21c4a8a5bece220a24dc28d1cbb5d1e5875be4a575d5d1da172a"
+dependencies = [
+ "crossterm",
+ "edtui-jagged",
+ "enum_dispatch",
+ "ratatui-core",
+ "ratatui-widgets",
+ "unicode-width",
+]
+
+[[package]]
+name = "edtui-jagged"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


agavra (approved):
heh this is pretty cool! thanks @thiblahute

PR #435, opened 2026-06-26
Session Type Detection For Clipboard On Linux

### LINKED ISSUE #434: Clipboard export uses `xclip` before `wl-copy` in Wayland, causing silent no-op copy (Linux)
Hello! Thanks for sharing tuicr with us, loving it!

## Issue

On a Wayland session where both `xclip` and `wl-copy` are installed, `xclip` can exit successfully but still not update the clipboard that the Wayland desktop uses. Since `tuicr` treats that as success, it never tries `wl-copy` or the OSC 52 fallback.

### Environment

- Terminal: Ghostty
- Session: Wayland
- `tuicr`: `0.18.0`
- `WAYLAND_DISPLAY` is set
- `XDG_SESSION_TYPE=wayland`
- Both `xclip` and `wl-copy` are available

## Repro

1. Open `tuicr`
2. Add a review comment
3. Run `:wq`
4. Confirm Copy review to clipboard? with `Y`

### Expected

The exported review markdown is copied to the system clipboard.

### Actual

Nothing useful is copied to the clipboard. There is no visible error because the clipboard command appears to succeed.

## Suspected cause

`src/output/markdown.rs` tries `xclip` before `wl-copy`:

```rust
try_clipboard_cmd("xclip", &["-selection", "clipboard"], text)
    || try_clipboard_cmd("wl-copy", &[], text)
```

This seems related to the Linux clipboard fix in #372, but this environment exposes a backend ordering problem rather than the original arboard lifetime issue.

## Possible fix

Prefer `wl-copy` when `WAYLAND_DISPLAY` is set, then fall back to `xclip`.

Roughly:

```rust
if std::env::var_os("WAYLAND_DISPLAY").is_some() {
    try_clipboard_cmd("wl-copy", &[], text)
        || try_clipboard_cmd("xclip", &["-selection", "clipboard"], text)
} else {
    try_clipboard_cmd("xclip", &["-selection", "clipboard"], text)
        || try_clipboard_cmd("wl-copy", &[], text)
}
```

I'm happy to submit a PR for this if you’re open to that direction.

mward25:
# Before

Clipboard would always run `xclip`, then, `wl-clipboard` (though because `xclip` doesn't fail on Wayland, `wl-clipboard` would never actually run).

# After

On Linux, `tuicr` will detect if `XDG_SESSION_TYPE` is `X11` or `Wayland`, then choose the proper clipboard tool to use.

# Tests

 - Tested on Hyprland (Wayland).
 - Tested on Openbox (X11).

 > Closes #434 
 
EDIT: Added Closes
EDIT: Added 'Tests' section

### DIFF
--- a/src/output/markdown.rs
+++ b/src/output/markdown.rs
@@ -96,8 +96,20 @@ pub fn copy_text_to_clipboard(text: &str) -> Result<bool> {
 
 /// Try xclip (X11) then wl-copy (Wayland). Returns true if either succeeds.
 fn try_copy_via_subprocess(text: &str) -> bool {
-    try_clipboard_cmd("xclip", &["-selection", "clipboard"], text)
-        || try_clipboard_cmd("wl-copy", &[], text)
+    let session = std::env::var("XDG_SESSION_TYPE");
+    if session.is_err() {
+        // Session not specified
+        return false;
+    }
+    let session = session.unwrap();
+    if session == "wayland" {
+        try_clipboard_cmd("wl-copy", &[], text)
+    } else if session == "x11" {
+        try_clipboard_cmd("xclip", &["-selection", "clipboard"], text)
+    } else {
+        // Session type unsupported
+        false
+    }
 }
 
 /// Try copying via a CLI tool that forks into the background and holds


agavra (approved):

agavra:
Thanks @mward25!

PR #442, opened 2026-07-02
fix(comments): associate comments with their commit when reviewing per-commit

thiblahute:
Comments were keyed only by (path, line), so when reviewing commit-by-commit via the inline selector, a comment made on commit A's line 42 also showed on commit B's line 42 — mixing up the diff view, comment navigator, and submit preflight.

- Add `commit_id: Option<String>` to `Comment` (serde `#[default]`, so legacy session JSON rehydrates as `None` = always visible), stamped when the inline selector shows exactly one commit.
- Visibility rule (`comment_visible_with`): visible if `commit_id` is `None` or its SHA is in the selected commit set. Applied in `rebuild_annotations`, both diff renderers, `file_render_body_height`, the comment navigator, and the submit preflight; hot paths compute the commit set once and pass it down.
- `find_comment_at_cursor` guards against stale annotations during an async commit-selection reload, so delete/edit no-op on hidden comments.
- Also fixes a latent bug: `delete_comment_at_cursor` and `enter_edit_mode` re-derived the comment index by counting side-matching comments, but `AnnotatedLine::LineComment.comment_idx` is the absolute index into the stored Vec. Both now use the absolute index directly.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -240,6 +240,8 @@ These are non-obvious things the implementation chain hit. Worth preserving for 
 
 12. **`cd` into the worktree before running `cargo`.** `cargo` resolves `Cargo.toml` from `pwd`. Running gates from the wrong worktree silently exercises the wrong tree.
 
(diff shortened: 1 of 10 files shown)


agavra:
Thanks! This makes sense to me.

PR #444, opened 2026-07-03
Add per-commit review mode

ruifm:
Adds an initial per-commit review workflow for local commit ranges. Reviewers can run `tuicr --commits -r main..HEAD`, move through commits one at a time, leave comments on each commit’s diff or commit message, and export feedback grouped by commit for agent-friendly review output.

Example export:

```
## Local tuicr Comments

### Commit abc1234 - Refactor auth token parsing

Author: Jane Developer

Commit message:

```text
Refactor auth token parsing

Move token parsing into a dedicated helper before changing validation behavior.

1. [SUGGESTION] commit message:1 - Make the subject imperative.
2. [NOTE] src/auth.rs:42 - This split makes the follow-up behavior change easier to review.

### Commit def5678 - Reject expired auth tokens

Author: Jane Developer

Commit message:

Reject expired auth tokens

3. [ISSUE] src/auth.rs:88 - This should handle missing expiry claims explicitly.
```

Advantages of per-commit mode:

- Keeps feedback attached to the exact commit that introduced the code.
- Makes commit hygiene review practical, including comments on commit messages.
- Helps reviewers evaluate whether each commit is internally coherent and reviewable.
- Produces export output grouped by commit, which is easier for authors and coding agents to
apply without mixing feedback across unrelated changes.
- Still supports normal line, range, file, and review-level comments, but scoped to the
active commit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -65,7 +65,7 @@ src/
 │   ├── mod.rs
 │   ├── comment.rs       # Comment, CommentType (Note/Suggestion/Issue/Praise)
 │   ├── diff_types.rs    # DiffFile, DiffHunk, DiffLine, FileStatus, LineOrigin
-│   └── review.rs        # ReviewSession, FileReview (the persisted review state)
+│   └── review.rs        # ReviewSession, CommitReview, FileReview (persisted review state)
 │
 ├── input/
 │   ├── mod.rs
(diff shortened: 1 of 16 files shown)


thiblahute:
We already have a very similar workflow and I added https://github.com/agavra/tuicr/pull/442 yesterday which is very similar to what you are doing, maybe you approach is better though.

(Not that I have been using tuicr to do per commit review for a long time)

agavra:
Thanks @thiblahute and @ruifm - I prefer the approach in #442 that fixes the bugs with the existing per-commit reviews rather than building workflow into `tuicr` itself. Workflows are highly personal, and you could code this up in userspace IIUC.

PR #448, opened 2026-07-05
feat(ui): detect light/dark theme from terminal background

veeti:
**TL;DR** This should fix the default theme appearing unreadable on systems with a light theme and dark terminal:

Before
image

tuicr 1

After
image

tuicr 2

YMMV but I thought this might be helpful.

----

Prefer to query the terminal background color using terminal-colorsaurus, then fall back to the system theme.

terminal-colorsaurus was chosen due to its minimal added transitive dependencies, comprehensive documentation and Windows Terminal support.

Seems to fix light system theme & dark terminal rendering unreadable text on KDE + Konsole, Windows 11 + Windows Terminal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2451,6 +2451,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "terminal-colorsaurus"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


agavra (approved):
hopefully 0.5 is an acceptable threshold, but this makes sense to me!

PR #450, opened 2026-07-08
fix(ui): make word wrap work in side-by-side diff view

### LINKED ISSUE #441: Word wrap has no effect in side-by-side diff view
`wrap = true` (and `:set wrap!`) works in unified view but is silently a no-op in side-by-side. Long lines are truncated with `...` regardless of the setting.

## Reproduce

1. Open a diff with a line longer than half the terminal width.
2. `:diff` to switch to side-by-side.
3. `:set wrap!` — no visible change; line stays truncated.
4. `:diff` back to unified — same line now wraps.

## Cause

In side-by-side, each column is hard-truncated/padded to a fixed `content_width` before being joined with `" │ "` into one `Line`. The joined line already fits the viewport exactly, so the subsequent `Paragraph::wrap` has nothing to wrap.

Unified view never truncates — it hands natural-width `Line`s to ratatui and lets `Wrap { trim: false }` do the work.

yuhua99:
Fixes #441.

Sorry for the delay on this one — work has been busy the past while, so
it took me a bit to get the PR out.

## Fix

Introduce a span-level word wrapper (`wrap_spans`) and route both views through it, so the same code that renders rows also produces their heights. Side-by-side additionally wraps each column independently and keeps the `│` divider aligned on continuation rows.

## Commits

1. **feat(ui): add span-level word wrapper wrap_spans**:
   style-preserving word wrapper primitive; no callers yet.
2. **refactor(ui): share wrap-aware cursor-line highlight across diff views**:
   extract a common `paint_cursor_line_highlight` that walks `row_heights` so unified and side-by-side can't drift on wrapped rows.
3. **refactor(ui): simplify side-by-side pair loop and expanded-context helper**:
   some cleanup to make room for the meta insertion in #5.
4. **fix(ui): single-pass wrap in unified view so row heights match rendered rows**:
   replace `Paragraph::wrap` + `compute_row_heights` dual pass with one `wrap_spans` call producing both the visual rows and their heights.
5. **feat(ui): per-column word wrap in side-by-side view**: 
   the actual fix. Record `SbsRowMeta` (raw content, prefix, pad style) per logical line at build time, then wrap each column at `content_width` with `│`-preserving continuation prefixes. Deletes the now-unused `compute_row_heights`.
6. **fix(ui): extend comment-box gutter bar over wrapped rows**:
   draw the comment-box vertical bar across every wrapped sub-row instead of stopping at the first visual row.
7. **feat(ui): replace front-draining with cursor walk in wrap_spans hard split**:
   perf. make it faster

### DIFF
--- a/src/ui/diff_side_by_side.rs
+++ b/src/ui/diff_side_by_side.rs
@@ -3,14 +3,14 @@ use ratatui::{
     layout::Rect,
     style::{Modifier, Style},
     text::{Line, Span},
-    widgets::{Block, Borders, Paragraph, Wrap},
+    widgets::{Block, Borders, Paragraph},
 };
 use unicode_width::UnicodeWidthStr;
 
 use crate::app::{
     App, DiffSource, ExpandDirection, FocusedPanel, GAP_EXPAND_BATCH, GapId, InputMode,
 };
-use crate::model::{FileStatus, LineOrigin, LineRange, LineSide};
+use crate::model::{DiffLine, FileStatus, LineOrigin, LineRange, LineSide};
 use crate::theme::Theme;
 use crate::ui::comment_panel;
 use crate::ui::diff_view::{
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thanks @yuhua99 this is a clear improvement. Some things don't work great (like `zz` doesn't always compute the correct center) though we can file those as follow ups as we figure it out.

PR #453, opened 2026-07-09
fix :edit on PR-s where you're forked

### LINKED ISSUE #451: :edit doesn't work when working on a fork-pr
I was making a classic github fork -> push -> PR to original contribution. Locally that means I have a remote `origin` for the upstream and `myfork` for, well, my fork. So I'm on main, one commit ahead of `origin/HEAD` and when I tried to call `:edit` on a file I got `Cannot open pkg/tui/views/adfmd.go: no local checkout`. 

I have a feeling this should be resolvable.

ferdinandyb:
This series fixes on issue where the `:edit` fails with no local checkout if you're operating on a fork-pr model.

Fixes: #451

### DIFF
--- a/src/forge/mod.rs
+++ b/src/forge/mod.rs
@@ -95,40 +95,31 @@ fn remote_urls(repo_root: &Path) -> Vec<String> {
     all_urls
 }
 
-/// Detect the forge repository for the local checkout at `repo_root`.
+/// Parse `url` as a forge remote repository.
 ///
-/// Tries GitLab first (host must contain "gitlab"); falls back to GitHub,
-/// which accepts any host so GitHub Enterprise remotes whose hostname does
-/// not literally contain "github" are still detected. Returns `None` when
-/// no remote can be parsed.
-pub fn detect_forge_repository(repo_root: &Path) -> Option<ForgeRepository> {
-    let all_urls = remote_urls(repo_root);
+/// Tries GitLab first — its parser already filters to "gitlab" hosts, so
+/// trying it first won't claim GitHub Enterprise remotes — then falls back
+/// to GitHub, which accepts any host (covers github.com and GHE hosts whose
+/// hostname does not literally contain "github").
+fn parse_any_remote_url(url: &str) -> Option<ForgeRepository> {
+    parse_gitlab_remote_url(url).or_else(|| parse_github_remote_url(url))
+}
 
-    // GitLab parser already filters to "gitlab" hosts, so trying it first
-    // won't claim GitHub Enterprise remotes.
-    for url in &all_urls {
-        if let Some(parsed) = parse_gitlab_remote_url(url) {
-            return Some(parsed);
-        }
-    }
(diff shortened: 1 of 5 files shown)


ferdinandyb:
Just a note: I'm very much NOT a rust dev, so AI was used. I'm pretty sure I understand what the code does. I'm not entirely sure about tests. Also "works on my machine".

src/forge/mod.rs
```diff
@@ -112,3 +120,72 @@ pub fn detect_forge_repository(repo_root: &Path) -> Option<ForgeRepository> {
     }
     None
 }
+
+/// `root`'s local checkout, but only when one of its remotes — not
+/// necessarily `origin` — matches `target_repo`.
+pub fn local_checkout_for_repo(root: &Path, target_repo: &ForgeRepository) -> Option<PathBuf> {
+    let matches = remote_urls(root).iter().any(|url| {
+        parse_gitlab_remote_url(url).as_ref() == Some(target_repo)
+            || parse_github_remote_url(url).as_ref() == Some(target_repo)
```
agavra: we should probably reuse `detect_forge_repository` or create some helper method so that if we add support for other git servers this won't need to be updated

agavra (commented):
I admit that `:edit` on a PR review (not a local patch) is a bit of a can of worms... you have to make sure that your local branch is exactly aligned with what's in the remote since I think we just fetch the diff instead of making the clone.

What's your workflow for this? We might want to disable `:edit` on PR reviews altogether or have some mode that fully checks out the branch for you to edit.

ferdinandyb:
> I admit that `:edit` on a PR review (not a local patch) is a bit of a can of worms... you have to make sure that your local branch is exactly aligned with what's in the remote since I think we just fetch the diff instead of making the clone.
> 
> What's your workflow for this? We might want to disable `:edit` on PR reviews altogether or have some mode that fully checks out the branch for you to edit.

So in my head `:edit` is "open this file in a proper editor so I can explore the context better". I agree you need to pay attention to the branch, but I don't think that is terrible. It's rather trivial to switch in the editor if need be, and actually when reviewing, if you need extra context it will be about origin/HEAD, not the current change-set. I mean the PR itself includes all the changes, so that should be trivially available in `tuicr`, for anything extra, the status quo should be enough.

src/forge/mod.rs
ferdinandyb: fair point

agavra:
ah, your use case makes a lot of sense to me. I wonder if we should be opening it in "read only" mode if the default editor has a way to do that (like VIM does). Anyway, we can leave that for a follow up happy to merge this one after addressing the comment

agavra:
also I apologize but https://github.com/agavra/tuicr/pull/456 was long overdue, so there'll be some conflicts. claude shoudl be able to handle the rebase

src/forge/mod.rs
ferdinandyb: Done

agavra (approved):
Thanks for the contribution @ferdinandyb :)

PR #454, opened 2026-07-09
Fix tabify context lines

### LINKED ISSUE #452: indentation seems to be stripped in the non-hunk part of the code
It looks like all indentation is stripped from the "expandable" areas:

So here's an edit in neovim:

Image

This is how it looks like in tuicr:

Image

And if I expand the context above:

Image

ferdinandyb:
Only the actual diffs are sanitized for tabs, not the context. Files
which use tabs are thus not indented at all in the context lines. Tabify
context lines as well.

Fixes: #452


agavra (approved):
nice catch and thanks for the cleanup!

PR #459, opened 2026-07-10
Fix false-positive binary detection on hunk headers containing "Binary"

antonio2368:
`is_binary_patch_line` matched any line containing "Binary", including `@@` hunk headers whose context text names a function like `deserializeBinaryBulkStatePrefix`. Now it requires the real marker shape (`Binary file...`/`GIT binary patch`) instead.


agavra (approved):
Derp! Thanks for the catch @antonio2368

Issue #427, opened 2026-06-22
Quitting a review session and returning to the PR main menu


tbloemen:
I might be missing a keybinding from the docs, but I could not find it. When you are in a PR, there is no way to quit from that pr and return to the main menu, where you can see the git commits and have an overview of the pull requests. If you press `esc` you just close the file sidebar and if you press `q` you exit the application alltogether. Same goes for `:wq`, `:x`, or `ZZ`. Is there a way to return to the main menu?

agavra:
You can use `:commits` to go back to the commit selection screen. Is that what you were looking for?

tadone:
I use `:prs` but when first starting to use tuicr I expected ESC to bring me back to the main section. Would be great to have shortcut to go back vs typing in the command.

agavra:
`esc` now toggles the menus, 🤔 open to suggestions for what a good shortcut would be.

tbloemen:
I think `esc` would still be the correct shortcut. I would then propose `tab` to be the toggle for opening and closing the sidebar, instead of only opening. `esc` would take you back to the commits/pr menu (whichever was last open)

agavra:
I'm in support of that @tbloemen - LMK if you want to submit that as a PR, otherwise I can add it to my backlog.

PR #469, opened 2026-07-15
fix: stop Esc from hiding the file panel, fixes #467

### LINKED ISSUE #467: Pressing escape repeatedly unexpectedly hides the file panel
When I open up tuicr, I get the default view.

Image

If I press "Escape" twice on the editor (e.g. after inserting a comment), the file panel goes away.

Image

I can type `;e` to bring it back, but it's not obvious.

Suggest that pressing escape shouldn't hide the file panel -`;e` to toggle it should remain.

a-h:
Esc in a normal-mode panel now only returns focus to the diff; it no
longer hides the file list. Visibility stays controlled by the explicit
toggles ;e / :e / <leader>e.

The hide-on-Esc behavior came from #373 (ADR FEAT-0016), which framed
Esc as "back to the diff, quiet the chrome". That assumed the file list
was transient chrome, but it has always defaulted to visible, so Esc
after leaving a comment box silently removed a panel the user never
asked to hide, with no discoverable way back.

### DIFF
--- a/src/app/tests/file_list_visibility_tests.rs
+++ b/src/app/tests/file_list_visibility_tests.rs
@@ -1,0 +1,129 @@
+use crate::app::*;
+use crate::handler::{handle_diff_action, handle_file_list_action};
+use crate::input::Action;
+use crate::model::FileStatus;
+use crate::vcs::traits::VcsType;
+
+struct DummyVcs {
+    info: VcsInfo,
+}
+
+impl VcsBackend for DummyVcs {
+    fn info(&self) -> &VcsInfo {
+        &self.info
+    }
+
+    fn get_working_tree_diff(&self, _highlighter: &SyntaxHighlighter) -> Result<Vec<DiffFile>> {
+        Err(TuicrError::NoChanges)
+    }
+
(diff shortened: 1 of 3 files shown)


agavra:
This makes sense to me, I don't love the current `esc` behavior (see also #427). What do you think about removing the current `<esc>` behavior and instead making it return to the commit selector? 

i model the `;e` etc... after NERDTree shortcuts, and `esc` doesn't return you back to the main open file either there.

a-h:
Oh, that makes sense. Going to the root menu option.

It's a separate thing, but I often exit tuicr and start the whole review again, especially if I'm doing `tuicr -A` where I need to stage some new files before they'll get picked up. I like to review code before each commit, but sometimes I just want to review everything that's in the directory, so "review the whole tree" regardless of staging/dirtiness, and/or commit history, so if that was in the list, that would be a great workflow.

One thing is that if you escape out to the root menu option, and haven't saved, I think you'd expect that if you selected the same commit or unstaged etc. you wouldn't lose your review, so would probably need the "save review" option to come up.

PR #470, opened 2026-07-15
fix: users can set diff.mnemonicPrefix = true in their git config which breaks diff parsing, override this setting in tuicr

a-h:
I noticed that the tests were failing for me.

This turned out to be a config setting I had in git. This change sets the config setting to be the one tuicr expects, overriding user settings.


agavra (approved):
TIL! I didn't know that this was a setting. thanks for the catch @a-h

PR #471, opened 2026-07-16
diff: render commit message as prose, not a diff

thiblahute:
In side-by-side view, commit message lines were duplicated across both columns and shown with per-column line numbers. This renders them as a single full-width, left-aligned, neutrally-styled block (like delta), and drops the line-number gutter for commit messages in the unified view too, since a commit message is prose, not code.


agavra (approved):
good stuff, patch makes sense

PR #473, opened 2026-07-16
feat: add installation-aware updates and rollbacks

pbjorklund:
Adds `tuicr update [VERSION]` so users can update the active installation or install a known-good version without first identifying how tuicr was installed. Homebrew, Cargo, Mise, and Nix keep ownership of managed binaries, while install-script and manually downloaded binaries use SHA-256-verified GitHub release assets and guarded in-place replacement. Exact versions support both upgrades and rollbacks for Cargo and direct binaries; Homebrew, Mise, and Nix return a clear error and leave pinning to their own workflows.

The test matrix covers the supported upgrade and rollback behavior for every documented installation method, then runs live direct-binary upgrade and rollback checks on Linux, macOS, and Windows. This crosses package-manager detection, release metadata, archive extraction, and executable replacement, so hopefully CI will catch any issues we missed.


src/update/check.rs
```diff
@@ -0,0 +1,125 @@
+use std::time::Duration;
+
+use ureq::Agent;
+
+const CRATES_IO_URL: &str = "https://crates.io/api/v1/crates/tuicr";
```
pbjorklund: Should this be hardcoded like this? What is repo convention?

src/update/check.rs
```diff
@@ -0,0 +1,125 @@
...
+        .is_some_and(|(latest_version, current_version)| latest_version > current_version)
+}
+
+fn parse_version(version: &str) -> Option<(u32, u32, u32)> {
+    let mut parts = version.split('.');
+    let major = parts.next()?.parse().ok()?;
+    let minor = parts.next()?.parse().ok()?;
+    let patch = parts.next().unwrap_or("0").parse().ok()?;
+    Some((major, minor, patch))
+}
```
pbjorklund: Seems brittle.

src/update/install.rs
```diff
@@ -0,0 +1,244 @@
+mod archive;
```
pbjorklund, line 1: File-level: I wonder if we should accept a param to install a specific version. Would a) be good for bad releases for a high agentic project like this and b) could use that in automated tests for the feature itself

src/update/install.rs
```diff
@@ -0,0 +1,244 @@
...
+use ureq::Agent;
+
+use self::archive::{extract_binary, release_asset_name, release_asset_url};
+use self::method::{detect_install_method, manager_command};
+use self::replace::replace_executable;
+use super::check::is_newer_version;
+
+pub use self::method::InstallMethod;
+
+const RELEASE_API_URL: &str = "https://api.github.com/repos/agavra/tuicr/releases/latest";
```
pbjorklund: Check repo convention and see how to handle config like this.

src/update/install.rs
```diff
@@ -0,0 +1,244 @@
...
+    CurrentExecutable(#[source] std::io::Error),
+    #[error("failed to fetch update metadata: {0}")]
+    Network(String),
+    #[error("invalid GitHub release metadata: {0}")]
+    ReleaseMetadata(String),
+    #[error("tuicr does not publish pre-built binaries for {arch}-{os}")]
+    UnsupportedPlatform { os: String, arch: String },
+    #[error("release asset '{0}' was not found")]
+    MissingAsset(String),
+    #[error("release asset '{0}' has no SHA-256 digest; refusing an unverified update")]
```
pbjorklund, line 71: Is this always there? How do we know? Documented?

src/update/install.rs
```diff
@@ -0,0 +1,244 @@
...
+        Ok(Self {
+            executable: std::env::current_exe().map_err(UpdateError::CurrentExecutable)?,
+            home: directories::BaseDirs::new().map(|dirs| dirs.home_dir().to_path_buf()),
+            cargo_home: std::env::var_os("CARGO_HOME").map(PathBuf::from),
+            current_version: env!("CARGO_PKG_VERSION").to_string(),
+            os: std::env::consts::OS.to_string(),
+            arch: std::env::consts::ARCH.to_string(),
+        })
+    }
+}
```
pbjorklund, line 121: This looks bad. Find alternatives that are easier to read and reason about

src/update/install.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    let binary = extract_binary(&asset.name, &archive)?;
+    runtime.replace_executable(&context.executable, &binary)?;
+    Ok(UpdateOutcome::Updated {
+        method,
+        previous_version: context.current_version.clone(),
+        new_version: latest_version.to_string(),
+    })
+}
```
pbjorklund, line 294: Research if anyone has a crate to handle this scenario already

src/update/install/archive.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::fmt;
+use std::io::{Cursor, Read};
+use std::path::Path;
+
+use flate2::read::GzDecoder;
+
+use super::UpdateError;
+
+pub(super) fn release_asset_url(version: &str, asset_name: &str) -> String {
+    format!("https://github.com/agavra/tuicr/releases/download/v{version}/{asset_name}")
```
pbjorklund: Good idea to hardcode?

src/update/install/archive.rs
```diff
@@ -0,0 +1,93 @@
...
+        _ => {
+            return Err(UpdateError::UnsupportedPlatform {
+                os: os.to_string(),
+                arch: arch.to_string(),
+            });
+        }
+    };
+    let extension = if os == "windows" { "zip" } else { "tar.gz" };
+    Ok(format!("tuicr-{version}-{target}.{extension}"))
+}
```
pbjorklund: Seems like the release tagets is shared responsibility and should not live here

src/update/install/archive.rs
```diff
@@ -0,0 +1,93 @@
...
+                .read_to_end(&mut binary)
+                .map_err(|error| archive_error(asset_name, error))?;
+            return Ok(binary);
+        }
+    }
+    Err(archive_error(
+        asset_name,
+        "archive does not contain the tuicr binary",
+    ))
+}
```
pbjorklund, line 34: Seems convoluted. same with zip

src/update/install/installation.rs
```diff
@@ -0,0 +1,95 @@
+use std::fmt;
```
pbjorklund, line 1: File-level: bad name for file

src/update/install/executable_swap.rs
```diff
@@ -0,0 +1,72 @@
+use std::fs::{self, OpenOptions};
```
pbjorklund: File-level: bad name for file

pbjorklund: File-level: needs heavy refactor to be more readable, more tests and guards since touchinh filesystem in this way

src/update/check.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Updated this to derive the crate path from Cargo's `CARGO_PKG_NAME` metadata, which matches the existing use of `CARGO_PKG_VERSION`. Only the external crates.io API base remains constant. Added a focused URL-construction test; the update-check tests, formatting, and Clippy all pass.

src/update/check.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Agreed. Replaced the handwritten parser with `semver::Version`, which handles prereleases and rejects incomplete or invalid versions according to SemVer. `semver` was already transitive, and is now declared directly. Update-check and installer tests pass, along with formatting and Clippy.

src/update/install.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Added `tuicr update [VERSION]` with SemVer validation. Direct installs can now install or roll back to an exact tagged release with the same SHA-256 verification, and Cargo installs pass the exact version to `cargo install --version`. Homebrew, Mise, and Nix return a clear pinning-workflow error rather than bypassing package-manager ownership. Added CLI, downgrade, command-delegation, metadata-mismatch, and URL tests; all 1,105 tests, formatting, Clippy, CLI help/error checks, and CodeGraph sync pass.

src/update/install.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Updated this to follow the existing Cargo-metadata convention used for package name and version. GitHub API and release-asset URLs are now centralized and derived from `CARGO_PKG_REPOSITORY`; no production updater URL embeds `agavra/tuicr`. All 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

src/update/install.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

GitHub has automatically computed SHA-256 digests for uploaded release assets since June 3, 2025, per the linked GitHub changelog now cited beside the field. I checked all 124 assets across tuicr's published releases; none lack a digest. The field remains optional so older or incomplete API data produces the existing fail-closed error instead of a deserialization failure. The missing/bad-digest test, formatting, Clippy, and diff checks pass.

src/update/install.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Refactored this so system discovery uses named steps and resolves the install method once. `UpdateContext` now carries the meaningful `InstallMethod` instead of leaking home-directory and `CARGO_HOME` detection inputs into the rest of the update flow. Added a context-level detection assertion; all 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

src/update/install.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Researched the available crates. `self_update` is the closest fit and handles GitHub releases, exact tags, tar/zip extraction, and cross-platform replacement. Its stable `0.42` release does not consume GitHub asset digests; the digest-aware implementation is only in `1.0.0-rc.6` and explicitly skips verification when a digest is absent, while this updater fails closed. `axoupdater` requires cargo-dist install receipts, which tuicr does not produce, and `self-replace` covers only the final replacement step. I kept the current focused implementation rather than weaken verification or add a release-candidate dependency; revisiting `self_update` after 1.0 stabilizes and supports required digests is the best path. No code changed; the worktree is clean.

src/update/install/archive.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Agreed; this was already addressed in `fb22e6d`. The outdated hardcoded URL was removed from `archive.rs`; release API and download URLs now live in `source.rs` and derive from Cargo's `CARGO_PKG_REPOSITORY`. The focused URL/target test passes, and no production updater code contains the old hardcoded release URL.

src/update/install/archive.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Moved release-target and asset-name selection into `source.rs`, alongside release repository/API/download concerns. `archive.rs` now only detects archive formats and extracts the binary. All 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

src/update/install/archive.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Simplified both paths around the archive contract enforced by the release workflow: tar reads the single root `tuicr` entry, ZIP uses `by_name("tuicr.exe")`, and both share one reader helper. Nested same-name entries are now rejected instead of searched by basename. All 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

src/update/install/installation.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Renamed `method.rs` to `installation.rs`. The module owns install-method detection and package-manager delegation, so the new name describes its responsibility rather than its implementation shape. All 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

src/update/install/executable_swap.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Renamed this to `executable_swap.rs`: the module stages new executable bytes, preserves the existing Unix mode, and swaps the staged file into place with Windows rollback. Renamed its internal entry point and test to `swap_executable` as well. All 15 installer tests, formatting, Clippy, diff checks, and CodeGraph sync pass.

pbjorklund: > Generated by AI via gh-review-watch on PI.

Heavily refactored this boundary. Platform-specific rename/backup logic now delegates to the established `self-replace` crate. The local adapter only validates the current executable, rejects relative/missing/directory/symlink/non-running targets and empty payloads, stages through a secure sibling tempfile, preserves permissions, calls `sync_all`, and cleans up on failure. Expanded swap coverage from one test to four focused tests covering success, guards, symlinks, permissions, callback failure, and cleanup. All 1,108 tests, formatting, Clippy, diff checks, and CodeGraph sync pass; I also copied the built binary to a temp directory and successfully downgraded it in place from 0.19.1 to 0.19.0 with no staging leftovers.

src/update/install.rs
```diff
@@ -0,0 +1,299 @@
...
+    executable: PathBuf,
+    method: InstallMethod,
+    current_version: String,
+    os: String,
+    arch: String,
+}
+
+impl UpdateContext {
+    fn current() -> Result<Self, UpdateError> {
+        let executable = std::env::current_exe().map_err(UpdateError::CurrentExecutable)?;
```
agavra, line 107: This was a comment from codex, seems legit:

```
  - [P1] Canonicalize the executable before detecting its installer — /
    Users/agavra/dev/tuicr/src/update/install.rs:107-110
    On macOS, std::env::current_exe() preserves the invocation symlink, so
    a normal Homebrew launch reports /opt/homebrew/bin/tuicr rather than
    its Cellar target. This classifies the primary documented installation
    method as Direct; the subsequent replacement then rejects that same
    symlink in validate_target, making tuicr update fail. Mise and Nix
    profile symlinks can encounter the same issue, so canonicalize the
    path before both method detection and replacement.
```

pbjorklund: > Generated by AI via gh-review-watch on PI.

Canonicalized the current executable before install-method detection and before passing the path to direct replacement. This resolves Homebrew, Mise, and Nix invocation symlinks to their managed targets while retaining the swap guard against symlink targets.

Added a Unix regression test that invokes through a symlink and verifies the resolved Cellar path is detected as Homebrew.

Validation: `cargo test update::install::tests` (21 passed), `cargo fmt -- --check`.

agavra (approved):
Thanks @pbjorklund this is nice :)

PR #474, opened 2026-07-16
feat: add search to help popup

pbjorklund:
The in-app help requires manual scrolling to find a keybinding or command. This lets `/` reuse the existing search prompt while keeping help open, then uses `n` and `N` for later or earlier matches. Matching is case-insensitive and shares one traversal implementation with diff search. The change was reviewed through tuicr; `cargo test` and `cargo clippy --all-targets -- -D warnings` pass.


agavra (approved):
this is great! with this we can probably consider eventually adding a collapse/expand functionatlity to the help page and maybe expand the scope of the keybindings (I think not everything is there but I might be wrong)

PR #475, opened 2026-07-16
fix(forge): support PRs with over 300 changed files

pbjorklund:
tuicr cannot open pull requests with more than 300 changed files. GitHub stops serving the combined diff at that point, and patches large enough to hit the limit also made file and commit navigation slow.

This PR removes that limit and keeps large reviews usable.

## Approach

Normal PRs still use `gh pr diff`. The fallback runs only for GitHub's specific 300-file error.

When tuicr is running inside a checkout for the same repository, it checks that the current branch name and `HEAD` match the PR metadata exactly. If the base commit is also available, Git can build the merge-base diff from the local object database. The command compares two commits, so dirty files, staged changes, and untracked files are not included.

If any check fails, tuicr creates a blobless bare repository under the system temporary directory. It can borrow objects from the matching local repository, including from a linked worktree's common Git directory, then fetches the missing base and PR objects into the temporary repository. The temporary repository is deleted when the diff is ready.

External diff, textconv, and replace objects are disabled for locally generated diffs. Other `gh pr diff` errors are returned as before.

## Other approaches tried and ruled out

- **GitHub's pull-request files API:** This was the first fallback implementation. It requires rebuilding unified patches from per-file responses, omits some patch data, and has its own 3,000-file ceiling.
- **Always create a temporary clone:** This is correct and remains the final fallback, but it repeats network and disk work when the exact commits are already present. On the large test PR it made startup roughly twice as slow.
- **Fetch into the user's checkout:** This avoids another repository but changes local Git state during a read-only review. It also needs cleanup and recovery if the process stops partway through.
- **Create a linked worktree under `/tmp`:** A worktree is useful when files must be edited or tested. This path only needs a commit-to-commit diff. Adding a worktree still writes administrative state into the user's repository, can leave stale entries after interruption, and does not solve missing objects without a fetch.
- **Trust any matching checkout:** Repository name or branch name alone is not enough. Forks, shallow clones, stale branches, detached heads, and advanced base branches can otherwise produce the wrong diff.

The current hybrid keeps the common checked-out-PR case local and read-only. Cases that cannot prove they have the exact commits use the isolated fallback instead.

Large-review rendering keeps only visible rows plus overscan. File and hunk positions are cached, and parsing, syntax highlighting, range saves, and cleanup no longer block the input loop. This follows the background-work direction from #288 without exposing partially built diffs to the UI.

The real test case changes 592 files and produces a 408,004-line patch. With the PR branch checked out, startup app initialization dropped from about 16 seconds to 8.4 seconds and did not run `gh repo clone`. File navigation takes 10-11 ms, cached commit-range changes take about 6 ms, and the 400,000-line release benchmark averages 165 microseconds per navigation frame.

The fallback tests cover exact and stale PR heads, dirty and staged state isolation, detached heads, shallow clones missing the base, advanced base branches, linked worktrees, local object borrowing, no-checkout clones, merge-base behavior, external diff and textconv suppression, replace objects, cleanup, and the 301-file end-to-end case. Local formatting, Clippy, 1,106 tests, the release performance benchmark, CodeGraph sync, and the security review pass.


pbjorklund:
Hi @agavra, 

I'm pretty unsure if this approach is even a mediocre idea. Problem I want to solve is that I have a PR that have gone a little overboard that I want to review and I can't right now.

Noted the **never** comments about using local repo as source of truth - not sure how dead set on that you are though? Would be neat to be able to review PRs of any size since I enjoy the traceability it gives my team to work against the PR comments.

Will leave this as draft, let me know what you think in general before I put more time (and tokens) into it.

AGENTS.md
```diff
@@ -143,7 +144,7 @@ Repository-managed agent integrations:
 
 ### Important Implementation Details
 
-- **Infinite scroll**: All files rendered into one `Vec<Line>`, then sliced by `scroll_offset`
+- **Virtualized infinite scroll**: `line_annotations` indexes the full review, while each frame retains only the viewport plus bounded overscan in `DiffLineBuffer`. Per-file annotation ranges let renderers skip files outside the viewport.
```
pbjorklund: P2: This overstates what the renderers do. `file_annotation_ranges` drives `update_current_file_from_cursor()`, while the renderers still walk the diff and let `DiffLineBuffer` trim the visible spans. Please say navigation uses the per-file ranges, and make the same correction in item 14 below.

pbjorklund: > Generated by AI via gh-review-watch on PI.

Corrected both AGENTS.md claims. Rendering is now attributed to `DiffLineBuffer`, while `file_annotation_ranges` is described as supporting current-file navigation from the cursor. Verified both updated passages against `update_current_file_from_cursor()` and the unified/side-by-side renderers; the Markdown prose scan found only pre-existing style warnings elsewhere in the file.

src/app/tests/persistence_merge_tests.rs
```diff
@@ -42,6 +42,18 @@ fn file_comment_ids(session: &ReviewSession) -> Vec<String> {
         .collect()
 }
 
+#[test]
+fn queued_pr_selection_save_preserves_persisted_comments() {
+    let identity = test_session();
+    let mut persisted = identity.clone();
+    push_file_comment(&mut persisted, "external", "keep me");
+
+    let saved = App::session_with_pr_commit_selection(identity, Some(persisted), Some((2, 4)));
```
pbjorklund: P1: This only exercises the pure helper. It never drives `persist_pr_commit_selection_range()` or the debounced worker in `queue_pr_commit_selection_save()`, so a regression in the actual save path would still pass. Please add a test that calls the public persistence path and asserts an existing file comment survives the queued selection save.

src/app/tests/target_selector_tests.rs
```diff
@@ -951,6 +953,24 @@ fn should_preserve_hunk_marks_hidden_by_pr_range_diff() {
         .unwrap();
 
     assert!(app.session.is_hunk_reviewed(&path, &hidden_key));
+    assert!(app.range_diff_files.is_some());
```
pbjorklund, line 956: P1: This still only covers the synchronous seam. The async `poll_pr_range_reload_events()` path has the stale-result guard, and a late `compare` response after a second range toggle or PR switch could still overwrite the current diff if that branch regressed. Please add a test that delivers a mismatched `PrRangeReloadEvent::Done` and asserts the current range diff is untouched.

src/app/tests/persistence_merge_tests.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Replaced the helper-only test with a PR-mode `App` test that calls `persist_pr_commit_selection_range()`, waits for the debounced worker, reloads the saved session, and verifies the existing file comment survives alongside the new range. Test builds now carry the thread-local temporary reviews directory into the worker so the real queued path is deterministic. Verified with `cargo test app::tests::persistence_merge_tests`, `cargo check`, `cargo fmt --check`, and `git diff --check`.

src/app/navigation.rs
```diff
@@ -778,14 +747,12 @@ impl App {
     }
 
     pub(in crate::app) fn calculate_file_scroll_offset(&self, file_idx: usize) -> usize {
-        let mut offset = self.review_comments_render_height();
-        for (i, file) in self.diff_files.iter().enumerate() {
-            if i == file_idx {
-                break;
-            }
-            offset += self.effective_file_height(i, file);
+        if self.is_single_file_view {
```
pbjorklund: [P1] Returning 0 here makes single-file jumps land on the review-comments block above the file whenever comments exist, because those rows are not decoration. `jump_to_file`, `toggle_reviewed_for_file_idx`, and `reload_diff_files` all rely on this offset, so single-file mode stops snapping to the file body. Keep the current file render start here even in single-file view.

src/app/pr.rs
```diff
@@ -203,7 +251,7 @@ impl App {
         };
         self.session.commit_selection_range = value;
         self.session.updated_at = chrono::Utc::now();
-        let _ = self.save_current_session_merging_external();
+        self.queue_pr_commit_selection_save(value);
```
pbjorklund, line 278: [P1] This turns commit-selection persistence into a best-effort background write, but the worker waits 150ms and nothing flushes it on exit. A quick quit after `(` / `)` can drop the new range. Please keep this save synchronous, or add a flush before shutdown.

src/app/pr.rs
```diff
@@ -263,24 +315,57 @@ impl App {
...
+        self.expanded_top.clear();
+        self.expanded_bottom.clear();
+        self.file_line_count_cache = self.range_file_line_count_cache.take().unwrap_or_default();
+        // The cumulative PR was already registered and directory-sorted before
+        // it moved into `range_diff_files`. Restore its render index when safe;
+        // state changes while viewing a subset invalidate this cache.
+        let stale_render_index = if let Some(cache) = self.range_render_index_cache.take() {
+            let annotations = std::mem::replace(&mut self.line_annotations, cache.annotations);
+            let file_ranges =
+                std::mem::replace(&mut self.file_annotation_ranges, cache.file_ranges);
```
pbjorklund, line 378: [P1] This restore path blindly reuses `range_diff_files`, so a later PR reload at the same head can leave the cached full diff stale. After a narrowed review, toggling back to all commits will replay old files and annotations from before the reload. Invalidate or refresh the cache before taking it here.

src/app/tests/target_selector_tests.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Added an async poll-path test that installs a newer active range request, injects a mismatched `PrRangeReloadEvent::Done` from the prior range, and calls `poll_pr_range_reload_events()`. It verifies the displayed range diff is unchanged and the newer request remains active. Verified with the targeted cargo test, `cargo fmt --check`, and `git diff --check`.

src/app/navigation.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Removed the single-file zero-offset shortcut, so file jumps now use the current file annotation range and fall back to the rendered review-comment height when that range is not yet rebuilt. Added a regression test with review-level comments that verifies a single-file jump lands at the file start rather than the comment block. Verified all 17 single-file-view tests, `cargo check`, `cargo fmt --check`, and `git diff --check`.

src/app/pr.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Kept selection saves off the input path, retained the worker handle, and added `flush_pr_commit_selection_save()` to close the channel and join the worker. Normal shutdown now flushes before terminal teardown, so an immediate quit waits for the debounced save. Updated the persistence test to flush immediately after queuing and then verify the range and existing file comment on disk. Verified the persistence tests, `cargo check`, `cargo fmt --check`, and `git diff --check`.

src/app/pr.rs
pbjorklund: > Generated by AI via gh-review-watch on PI.

Invalidated `range_diff_files` and its line-count/render-index companions before installing a same-head reload result. Large cached vectors are moved to a background drop so reload handling stays bounded on the input thread. Added a regression test that seeds the old cache, reloads fresher same-head files, toggles back to all commits, and verifies the fresh diff remains. Verified all 75 target-selector tests, `cargo check`, `cargo fmt --check`, and `git diff --check`.

hydraxman:
Fresh reproduction on current `main` (`246f839e83ac4149489c43357e5ee2230bd16676`):

```text
$ gh pr diff 159789 -R rust-lang/rust --color never
could not find pull request diff: HTTP 406: Sorry, the diff exceeded the maximum number of files (300).
PullRequest.diff too_large
```

That PR currently reports 1,022 changed files. The pull-request files API returns ten full 100-file pages and a final 22-file page, which also confirms why that API would need bounded pagination and still retain its 3,000-file ceiling. I found no other open PR in this repository implementing the >300-file fallback; this remains the active implementation for #521.

pbjorklund:
@agavra any thoughts?

benja2998:
Not sure why you would want to review PRs that large. Shouldn't those PRs be split into multiple PRs?

agavra:
Thanks for the patience getting back to you guys on this one, I'm in agreement with @benja2998 I'm not super eager to add a bunch of complicated logic to support diffs with over 300 changed files. I doubt people are really reviewing those by hand anyway.

What's your use case for this?

pbjorklund:
@agavra most of our PRs are that large nowadays (monorepo with docs, code etc etc)

hydraxman:
The `hunk_positions()` change at `fe5675fb9b1ad87800923aa8d04997f401549a8b` also addresses #719: expanding context currently leaves `[` / `]` using raw diff-line offsets. I reproduced this independently on current main (`fbc3cfce`) and found the overlap while checking the final diff, so I am not opening a competing implementation PR.

[Four regression tests](https://github.com/hydraxman/tuicr/blob/f688b1999400d08f1577562782dd21723d108ba2/src/app/tests/expand_gap_tests.rs#L136-L353) are available to reuse. All four fail against current main and pass with the annotation-based target calculation:
- partial/full context expansion and collapse in both layouts and single/multi-file view;
- paired add/delete rows plus wrapped comments;
- actual TUI rendering after a hunk jump, including header visibility in a small viewport;
- asymmetric cross-file jumps with a hidden middle file.

I have not run these against this PR's full branch; the GREEN result applies to the narrow annotation-based fix on current main. The complete local test suite, build and `cargo clippy -- -D warnings` also pass there.

PR #477, opened 2026-07-17
fix(ui): make scroll and cursor jumps wrap-aware

yuhua99:
Follow up to #450.

That PR made word wrap real: one logical diff line can now span multiple
screen rows. However, all scroll/cursor math still assumed 1 logical line
equals 1 screen row.

With wrapped lines on screen:

* `zz`/`zt`/`zb` center on the wrong row
* `G` overshoots
* PageUp/PageDown skip or repeat content
* comment navigator jumps land the viewport in the wrong place

## Fix

Introduce `ui/row_height.rs`. `annotation_row_height(app, idx)` mirrors the
renderers' wrap pass to compute the exact visual row count of one annotation
without rendering.

All jump functions are rewritten on top of a single primitive,
`scroll_offset_for_rows_above`, which walks backward from an anchor spending
a screen row budget. Page keys now convert their row budget into a logical
line count first.

To keep the height mirror from drifting:

* every row format string is extracted into shared builders in
  `ui/diff_view.rs`, the single source for both renderers and the height
  oracle
* parity tests render into a `TestBackend` and compare observed per
  annotation heights
* a coverage assert forces every `AnnotatedLine` variant to be exercised

### DIFF
--- a/src/app/gaps.rs
+++ b/src/app/gaps.rs
@@ -1,7 +1,7 @@
 use super::*;
 
 impl App {
-    pub(in crate::app) fn gap_size(&self, gap_id: &GapId) -> Option<u32> {
+    pub(crate) fn gap_size(&self, gap_id: &GapId) -> Option<u32> {
         let file = self.diff_files.get(gap_id.file_idx)?;
 
         if gap_id.hunk_idx == file.hunks.len() {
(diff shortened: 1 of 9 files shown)


yuhua99:
Rebased onto latest main to resolve a conflict with #471

agavra:
Thanks @yuhua99 i fixed a few issues found while testing this + codex review and just pushed directly to speed things up. thanks for your patience on it! merging soon.

PR #484, opened 2026-07-19
commit selector: toggle key, hidden cycling, and order/selection options

thiblahute:
Improvements to the inline commit selector for multi-commit reviews:

- `<leader>s` toggles the pane (alongside `:set commits!`), and `( / )` keep cycling the selected commit while it is hidden.
- `show_commits = false` starts with the pane hidden, paralleling `show_file_list`.
- `commit_order = ascending` renders the pane oldest-first (presentation only; selection, persistence and since-last-review logic untouched).
- `commit_selection = oldest` opens the review scoped to the oldest commit for a walk-forward per-commit review.
- The status bar now shows the selected commit's position (e.g. `commit 2/3`) instead of a constant `1/N` when a single commit is selected.


agavra (approved):
all good changes, thanks again @thiblahute

PR #486, opened 2026-07-22
feat: add show_commits config to control commit selector visibility on startup

astromberg-wb:
## Description

Adds an optional `show_commits` config key mirroring `show_file_list`, controlling whether the commit selector panel is shown on startup.

- unset → current behavior: auto-shown for multi-commit reviews (>1 commit)
- false → starts collapsed
- true → starts open

Applied in the same startup config block as show_file_list in main.rs. Runtime toggles (`:set commits` / `:set nocommits` / `:set commits!`) are unchanged. Documented in `docs/CONFIG.md` with parse tests added.

## Rationale

I work on a 14 inch MacBook, and I typically review lots of commits at once. The file tree helps flow, the commit panel does not (I already know the scope I'm reviewing) and it takes valuable vertical space. I don't want to change the default setting, but I would love to have a config I can change so I don't have to run `:set nocommits` every time I open `tuicr`.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -27,6 +27,7 @@ theme_light = "gruvbox-light"
 diff_view = "side-by-side"
 ignore_whitespace = false
 show_file_list = true
+show_commits = true
 mouse = true
 leader = ","
 comment_vim = false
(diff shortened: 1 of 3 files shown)


agavra:
Ah sorry looks like #484 got there first and i merged that one. CLosing this in favor of that but LMK if I missed something that PR doesn't cover.

Thank anyway @astromberg-wb 🙏

Issue #468, opened 2026-07-15
Feature request: navigate comments globally


a-h:
When I've completed a review, I often want to check through the locations in a local review where I made comments, and validate that the changes I wanted have been done to my satisfaction.

Although the "Reviewed" status usefully resets, which lets me see which files have changed, I can't easily get through to the next comment in all files.

Looking in the help, there's a "comment navigator" section, but this section only shows up on a per-file basis, and it shows up at the bottom of the file list, so to get to it you have to use `;h` to go left, and `;j` to go down to select it, then you can navigator comments in just that file.

It's possible to delete all comments with `:clearc`, but not to iterate through and delete individual comments.

The workflow I'm looking for is:

- Show a list of all comments in the review
- Select a comment, which causes navigation to it
  -  The code around the comment since it may now be different due to the LLM changing the code based on the review comment, so I need to be able to scroll etc.
- Delete the comment with `dd` if it's resolved and move on to the next comment.

Then, I can revisit any review comments I don't feel are sufficiently handled.

ubuntudroid:
I also need this **during** a review, not just after changes have been made. Example flow:

1. Review file A and add a comment
2. Mark A as reviewed
3. Review file B and notice that it supersedes my comment in A
4. Want to jump back to my comment in A, but can only do so by moving back to file A, marking it as non-reviewed and then open the comment via cursor selection or via the now open comment navigator

A persistent comment navigator would make this much more comfortable.

a-h:
Possibly resolved in https://github.com/agavra/tuicr/commit/3b8575d14b9c571efbb1ea287ef6d2c4f7d3fe8b

a-h:
Confirm closed by #489

PR #489, opened 2026-07-24
Add diff comment navigation

cpaasch-oai:
Allow normal mode to jump between rendered comments with m/M. The cycler walks review, file, and line comments in display order, wraps at the ends, and keeps multiline comments as a single target.

Document the keybindings in README, help, and status hints, and cover forward/backward cycling plus empty-review behavior in tests.


agavra (approved):
love it, that'll be really helpufl for remote reviews.

a-h:
Possibly closes #468

PR #492, opened 2026-07-27
Fix tree selection when collapsing directories

### LINKED ISSUE #57: collapsing a file directory in the browser moves cursor up
if you hit "enter" to collapse a part of the file tree it will move the selected file up by (i think) the number of files collapsed. we should maintain the position in the tree but go up to the collapsed root.

related to #50

Guflly:
Keep the selected directory row active when it is collapsed, even when the diff panel is showing a file elsewhere in the tree.

Adds regression coverage for the file-list selection behavior.

Fixes #57


agavra (approved):
Thanks @Guflly for the contribution!

PR #493, opened 2026-07-27
fix: keep Escape from hiding file panel

### LINKED ISSUE #467: Pressing escape repeatedly unexpectedly hides the file panel

Guflly:
Fixes #467.

Normal-mode Escape now stays unbound instead of dispatching `ExitMode`, so repeated Escape presses no longer hide the file panel. Escape still exits the other modes as before.

Tested with:

- `cargo test should_leave_escape_unbound_in_normal_mode`
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets`

The full Windows run passed 1,130 tests. Its three existing failures are Windows path-separator and repository-ownership assumptions.

### DIFF
--- a/src/input/keybindings.rs
+++ b/src/input/keybindings.rs
@@ -213,7 +213,7 @@ fn map_normal_mode(key: KeyEvent, leader_key: char) -> Action {
         (KeyCode::Char(':'), _) => Action::EnterCommandMode,
         (KeyCode::Char('/'), _) => Action::EnterSearchMode,
         (KeyCode::Char('?'), _) => Action::ToggleHelp,
-        (KeyCode::Esc, KeyModifiers::NONE) => Action::ExitMode,
+        (KeyCode::Esc, KeyModifiers::NONE) => Action::None,
 
         // Quick quit
         (KeyCode::Char('q'), KeyModifiers::NONE) => Action::Quit,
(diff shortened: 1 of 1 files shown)


Guflly:
The update job hit a 403 while fetching release metadata. I don't have permission to rerun it—could someone rerun the workflow?

agavra:
I'm in support of this, I admit the `esc` behavior wasn't fully thought through when we added it. c.f. #469

src/input/keybindings.rs
```diff
@@ -213,7 +213,7 @@ fn map_normal_mode(key: KeyEvent, leader_key: char) -> Action {
         (KeyCode::Char(':'), _) => Action::EnterCommandMode,
         (KeyCode::Char('/'), _) => Action::EnterSearchMode,
         (KeyCode::Char('?'), _) => Action::ToggleHelp,
-        (KeyCode::Esc, KeyModifiers::NONE) => Action::ExitMode,
+        (KeyCode::Esc, KeyModifiers::NONE) => Action::None,
```
agavra, line 216: a bit weird to leave this here, but i'm fine having it be a placeholder for now. maybe a good beahvior going forward would be to have it return to the commit selector mode

agavra (approved):

PR #496, opened 2026-07-28
feat(export): make the review export preamble configurable

### LINKED ISSUE #487: [Feature Request] Make the exported review preamble configurable
### Summary

When you export a review (`y` to clipboard, `:clip`, or stdout), tuicr prepends a fixed preamble to the markdown. No `config.toml` key changes or suppresses it. I'd like the preamble to be configurable.

### Current behavior

`generate_markdown` emits the preamble unconditionally (`src/output/markdown.rs`, v0.19.1). Three hardcoded string literals make it up:

- Intro line (`src/output/markdown.rs:243-246`):
  > `I reviewed your code and have the following comments. Please address them.`
- Scope line, unstaged case (`src/output/markdown.rs:256-259`):
  > `Reviewing unstaged changes`
- Local comments header (`src/output/markdown.rs:394`):
  > `## Local tuicr Comments`

None of the recognized `config.toml` keys touch these. I checked `src/config/mod.rs`: `theme*`, `appearance`, `backend`, `comment_types`, `diff_view`, `export_legend`, and the rest. Nothing covers the export intro or headers.

### Motivation

- **Context hygiene**: this overlaps with the concern in #19 about keeping inline instructions short to avoid polluting the agent's context. A config key lets you trim or replace the preamble to fit your own harness.
- **Other consumers**: the output also feeds non-agent tooling and agents with their own review conventions, where the fixed intro framing adds noise.
- **No workaround for the clipboard path**: downstream filtering with `sed` works only for the stdout export. The interactive `y` yank writes straight to the clipboard with no hook in between (`src/output/markdown.rs:53-103`).

### Proposed solution

Add optional export keys that default to current behavior, so nothing changes for existing users. This follows the existing flat `export_legend` key:

```toml
# Override the intro line. Empty string omits it.
export_intro = "Code review comments:"

# Omit the "Reviewing <scope>" line.
export_scope_line = false

# Override the local comments header.
export_comments_header = "## Comments"
```

An `[export]` table would work equally well. Unknown or invalid values would warn and fall back to defaults, matching the current config-parsing behavior.

### Related

- #19: improve generated prompt output (overlaps on preamble content, from the agent-optimization angle)
- #149: other ways to get and use the output

ewhanhan:
Closes #487.

## What

Every exported review opened with a fixed intro line, a `Reviewing <scope>` line, and hardcoded section headings, with no config key to touch them. Filtering downstream with `sed` worked only for `--stdout`; the `y` and `:clip` clipboard path writes straight to the clipboard with no hook in between.

This adds an `[export]` section. Every key defaults to what tuicr already emitted, so exports stay byte-identical until you set one, and an empty string omits a line along with the blank line after it.

```toml
[export]
intro = ""
scope_line = false
comments_header = "## Comments"
```

| Key                      | Default                                                                      | Controls                                |
| ------------------------ | ---------------------------------------------------------------------------- | --------------------------------------- |
| `intro`                  | `I reviewed your code and have the following comments. Please address them.` | Opening line                            |
| `scope_line`             | `true`                                                                       | The `Reviewing <scope>` line            |
| `pr_metadata`            | `true`                                                                       | The `URL:` and `Head:` lines in PR mode |
| `comments_header`        | `## Local tuicr Comments`                                                    | Local comment heading                   |
| `remote_comments_header` | `## Existing GitHub Comments`                                                | Upstream thread heading                 |
| `legend`                 | `true`                                                                       | The `Comment types:` line               |

## Two calls worth your review

**`[export]` table instead of flat keys.** The issue proposed `export_intro`, `export_scope_line`, and so on. I mirrored `[forge]` instead, which gives scoped `export.<key>` typo warnings and somewhere for the existing `export_legend` to live. Happy to flatten it if you prefer the issue's shape.

**Defaults unchanged instead of dropping the preamble.** You mentioned agents have improved enough to omit it altogether. This ships the mechanism with defaults intact so nothing changes for existing users; flipping them is a small follow-up that stays independently revertible.

## Why `pr_metadata` is separate

The pull request branch emitted three lines, not one: the scope line plus `URL:` and `Head:`. Those two carry context an agent needs to fetch the PR, so folding them under `scope_line` would have stripped them along with the prose. They get their own key.

## Backward compatibility

`export_legend` still works. `[export].legend` wins when set. When `[export]` omits `legend`, `export_legend` stays in force, so adding a block to trim the intro does not switch the legend back on. That rule lives in `AppConfig::resolved_export` with tests for all four combinations.

**Breaking for library consumers:** `generate_export_content` and `export_to_clipboard` now take `&ExportConfig` in place of `show_legend: bool`. Both are public in the lib crate.

## Verification

- 1161 tests passing, 23 new
- `cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` clean
- The threading commit changed zero existing assertions, so the tests pinning the old preamble text still pass untouched
- Dogfooded per CONTRIBUTING with `intro = ""` and `scope_line = false`:

```
## Session: agavra/tuicr@ehan-feat-487/commits/5e9a1d7..5e9a1d7

## Local tuicr Comments

1. `AGENTS.md:301` (commit 5e9a1d7) - zz
```

## AI usage

Built with Claude Code (Opus 5), following the workflow in CONTRIBUTING.md. Every commit carries a `Co-authored-by: Claude` trailer.

Claude wrote the implementation, tests, and docs. I made the design calls: keep current defaults rather than dropping the preamble, and bring the remote comment header into scope. An adversarial review of the proposed design, run before any code was written, caught two defects that changed the final shape:

- Copying the `[forge]` parsing pattern would have reset `export_legend` to its default whenever an `[export]` block omitted `legend`. Fields are now `Option`-typed so absence stays distinguishable, with `AppConfig::resolved_export` covering the precedence.
- A single `scope_line` toggle would have stripped the PR `URL:` and `Head:` lines. They now sit behind their own `pr_metadata` key.

I ran the dogfood myself against a live review session in this repo; the output above is that run.

## Docs

`docs/CONFIG.md` gains an Export section, and `export_legend` is documented for the first time. `AGENTS.md`'s Review Format sample was corrected: it omitted the `Reviewing <scope>` line and the `## Local tuicr Comments` heading that tuicr already emitted, and it is now marked a default rather than a contract so agents parse the numbered list. `docs/CONFIG.md` was also missing from the Keeping Docs Updated table.

## Not in this PR

`remote_comments_header` defaults to `## Existing GitHub Comments` on GitLab merge requests too. This makes a workaround possible; the real fix is forge-aware defaulting, which I would file separately.


agavra (approved):
Thanks @ewhanhan :)

PR #497, opened 2026-07-28
feat(ui): filter the file tree by include/exclude regex

patrickjmccauley:
This is meant to add similar functionality to the vscode `files to include` + `files to exclude` filters when searching the codebase, though in a more targeted way using basic regex to filter in/out files based on path.

Large diffs can be hard to navigate when only part of the change is relevant - this PR adds include/exclude filtering plus a path search to the file tree, active while the tree is focused:

  `i`   include prompt — keep only paths matching a regex
  `e`   exclude prompt — drop paths matching a regex
  `I`   clear the include filter
  `E`   clear the exclude filter
  `/`   search paths, n/N step matches (wraps)

Patterns are case-insensitive and match the full relative path. Include runs first, then exclude removes from what is left.

A filter narrows the diff pane as well as the tree, so `i \.rs$` means you actually only see the Rust files: hidden files also drop out of `}`/`{` file navigation, `[`/`]` hunk navigation, and the file and +/- counts in the header. Implemented as a view over `diff_files` rather than a mutation of it — `file_idx` stays an absolute index, so nothing downstream needs remapping. `App::file_passes_filter` is the single predicate, consulted by `build_visible_items`, `rebuild_annotations`, both diff renderers, `hunk_positions`, and `file_render_height`/`effective_file_height`, which report zero lines for hidden files so cumulative scroll offsets stay correct.

Keys are focus-scoped through `map_file_tree_mode`: the tree claims i/e/I/E and `/`, while the diff keeps `i` for editing the comment at the cursor and `/` for searching the diff. An open prompt is a text-input sub-state of Normal mode, mirroring the target selector's `pr_filter_draft`.

Enter applies, Esc cancels, ctrl-u/ctrl-w edit the line. Reopening a prompt pre-fills the applied pattern, so submitting an emptied buffer is the same as I/E. An invalid regex reports the reason and leaves the prompt open to be fixed. The tree title reports what is hidden (`Files · 2/12 · 12 of 58`) and the bottom border shows the active patterns.

Filters are a view, not an edit: comments on hidden files are not deleted and still export. They are session-local — never persisted — but survive `:e`.

Screenshot 2026-07-28 at 10 02 27 AM

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,7 @@ src/
 ├── config/
 │   └── mod.rs           # User config loading (XDG on Unix, %APPDATA% on Windows)
 ├── app.rs               # Application state (App struct, InputMode, etc.)
+│   └── file_filter.rs   # File-tree include/exclude regex filters + `/` path search
 ├── error.rs             # Error types (TuicrError enum)
 ├── editor.rs            # External $EDITOR command construction and launch helpers
 ├── review_store.rs      # Library API for session listing/loading and shared comment insertion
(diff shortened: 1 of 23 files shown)


agavra (approved):
Thanks, this is a neat feature!

PR #501, opened 2026-07-28
fix(keybinds): remove support for arrow keys

benja2998:
While this software is clearly aimed at Vim users, it weirdly has support for arrow keys. Arrow keys make you a worse Vim user, and since this is targeted at Vim users the support for arrow keys should be removed.


benja2998:
I'm going to run cargo fmt.

agavra:
I appreciate the suggestion @benja2998 but I think many vim setups support arrow keys as well, i'm not sure what the downside is. Closing this PR, you're welcome to open an issue instead and we can leave it open to see if there is support for it.

PR #505, opened 2026-07-29
Feat/open editor with e

### LINKED ISSUE #482: Feature request: single-key binding to open the focused file in $EDITOR (like lazygit's `e`)
## Problem

Opening the file under the cursor in an external editor currently requires typing the full `:edit` command — and `:e` can't be used as a shorthand since it's taken by reload. Coming from tools like lazygit (where `e` edits the selected file), this is a frequent enough action during review that a command-mode round-trip feels heavy.

## Proposal

Bind `e` in the diff view (and possibly the file list) to open the focused file in `$EDITOR` — the same action as `:edit`. As far as I can tell from docs/KEYBINDINGS.md, `e` is currently unused in the diff view, and it matches the convention lazygit users already have in muscle memory.

Nice-to-have on top: jump to the cursor's line by invoking the editor with `+N` (most editors support it), so you land exactly where you were reviewing.

## Alternative

If adding fixed keys isn't desirable, user-configurable keybindings in `config.toml` would cover this and similar requests — though that's a much bigger feature; a built-in `e` binding would already solve the common case.

Thanks for tuicr — great tool!

ryuheechul:
Resolves #482.

## Summary

Adds a single-key `e` binding to open the focused file in `$EDITOR` (same action as `:edit`), and extends editor line-navigation support to Emacs.

### `e` key opens focused file in `$EDITOR`

- **`src/input/keybindings.rs`**: Adds `Action::EditFile` variant (under Review actions section) and maps unmodified `e` → `Action::EditFile` in normal mode.
- **`src/handler.rs`**: Dispatches `Action::EditFile` to `app.queue_editor_for_focused_item()` in `handle_shared_normal_action`, covering the diff view, file list, comment navigator, and inline commit selector.
- **`src/ui/help_popup.rs`**: Adds `e` help entry.
- **`docs/KEYBINDINGS.md`**, **`README.md`**: Document `e` binding.

Works from the diff view (also passes the cursor line as `+N` to supported editors) and from the file list, matching the lazygit `e` convention.

### Emacs `+NN` line navigation support

- **`src/editor.rs`**: Adds `emacs` and `emacsclient` to the `PlusLine` editor family so they receive `+42 /path/to/file` syntax instead of silently dropping the line number. `emacs -nw` and similar argument patterns are preserved correctly.

## Tests

- All existing lib tests pass.
- keybinding tests pass, including the updated `should_map_lowercase_e_to_edit_file_in_normal_mode`.
- editor tests pass, including new `emacs_receives_plus_line_before_path` and `emacs_args_are_preserved`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ A first-session cheatsheet. Press `?` inside tuicr for the full reference.
 | `v` / `V` | Visual mode (range comment) |
 | `r` | Toggle file reviewed |
 | `R` | Toggle hunk reviewed |
+| `e` | Open focused file in `$EDITOR` |
 | `y` | Copy review to clipboard |
 | `:edit` | Open focused file in `$EDITOR` |
 | `:submit` | Push review to GitHub or GitLab |
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -83,6 +83,7 @@ Shown below the file tree when local comments or visible remote PR threads exist
 | `dd` | Delete comment at cursor |
 | `i` | Edit comment at cursor (vim: text cursor at start) |
 | `A` | Edit comment at cursor with text cursor at end (vim mode only) |
+| `e` | Open focused file in `$EDITOR` |
 | `y` | Copy review to clipboard |
 
 ## Visual mode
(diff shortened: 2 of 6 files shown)


agavra (approved):
great addition, thanks for the contribution @ryuheechul ;)

ryuheechul:
> great addition, thanks for the contribution @ryuheechul ;)

You are welcome, @agavra!

PR #506, opened 2026-07-29
Strip SSH port before building forge --repo URLs

theherk:
Discovered while switching to the Pull Requests tab on the initial `tuicr` selector screen: the tab failed to load PRs against a self-hosted GitLab remote.

Root cause: `ssh://user@host:port/owner/repo` remotes (common for self-hosted GitLab/GHE on a non-default SSH port) leaked `host:port` into `ForgeRepository.host`. That host built the `--repo https://host:port/...` argument passed to `glab`/`gh`, which fails at the TLS layer since the SSH port isn't the HTTPS port.

Repro, confirmed against a real self-hosted GitLab remote on port 2222:
```
$ glab mr list --repo "https://gitlab.example.com:2222/owner/repo" ...
Get "https://gitlab.example.com:2222/api/v4/...": tls: first record does not look like a TLS handshake.
```

**Fix:** `strip_port()` added to both `parse_gitlab_remote_url` and `parse_github_remote_url`; drops the trailing `:<port>` before constructing `ForgeRepository`.

**Testing:** regression tests added for both parsers (`should_strip_ssh_port_from_self_hosted_ssh_scheme_remote`, `parses_ssh_remote_url_with_custom_port`). Full `cargo test --lib` passes.

### DIFF
--- a/src/forge/github/gh.rs
+++ b/src/forge/github/gh.rs
@@ -619,7 +619,7 @@ pub fn parse_github_remote_url(remote_url: &str) -> Option<ForgeRepository> {
         .map(|(_, rest)| rest)
         .unwrap_or(without_scheme);
     let (host, path) = without_user.split_once('/')?;
-    repository_from_path(host, path)
+    repository_from_path(strip_port(host), path)
 }
 
 fn parse_numeric_target(target: &str) -> Option<PullRequestTarget> {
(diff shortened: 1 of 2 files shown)


theherk:
I think the failing job is some transient network issue; rate limit probably.

agavra (approved):
Thanks! I appreciate any QA testing on self hosted instances since I don't have one to test against myself 🙏

PR #507, opened 2026-07-29
Disable commit signing in git/jj test fixtures

theherk:
Several `vcs::git` and `vcs::jj` tests create real commits in temp repos via the actual `git`/`jj` binaries. Contributors with commit signing configured globally (`commit.gpgsign = true`, jj's `signing.behavior = "own"`) get a real pinentry/GPG prompt on every `cargo test` run.

**Fix:**
- Git test helpers (`vcs/git/cli.rs`, `vcs/git/libgit2.rs`, `vcs/pristine.rs`) and the shared `run_git_command` (`vcs/git/mod.rs`) now pass `-c commit.gpgsign=false`.
- jj needs more care: `jj git init` signs its initial commit before any per-repo config override can apply, and the `JjBackend` methods under test call `run_jj_command` (production code, left untouched so real signing config keeps working for actual users), which also auto-snapshots and signs. Fix: point `JJ_CONFIG` at a throwaway config file, set once process-wide via `std::sync::Once`. Every `jj` child process spawned for the rest of the test run inherits the override, test-issued or production-issued alike, without touching the developer's real `~/.config/jj`.

**Testing:** full suite run with `GNUPGHOME` pointed at an empty, keyless directory. A real signing attempt would fail loudly ("No secret key") instead of passing silently, so the clean pass (1141 passed, 0 failed) proves no signing calls occur anywhere in the suite, regardless of a contributor's local config.

**Note on complexity:** the jj fix isn't trivial; a `Once`-guarded process-wide env var and a throwaway config file are more machinery than the git-side `-c` flag. That cost buys a real benefit for contributors who sign their own commits: no more surprise pinentry prompts during a plain `cargo test`. Worth the added surface area.

### DIFF
--- a/src/vcs/git/cli.rs
+++ b/src/vcs/git/cli.rs
@@ -1168,8 +1168,12 @@ mod tests {
     use crate::vcs::git::{diff, repository};
 
     fn git(workdir: &Path, args: &[&str]) {
+        // `-c commit.gpgsign=false` overrides any global signing config so
+        // contributors with commit signing enabled aren't prompted to sign
+        // throwaway commits in these temp repos.
         let output = Command::new("git")
             .current_dir(workdir)
+            .args(["-c", "commit.gpgsign=false"])
             .args(args)
             .output()
             .expect("failed to run git");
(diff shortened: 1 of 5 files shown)


agavra (approved):
makes sense to me, thanks @theherk

PR #510, opened 2026-07-29
fix(jj): remove ANSI escape sequences by disabling revision ids colo…

### LINKED ISSUE #508: `jj` integration breaks when color output is enabled
Hi! I think I found a bug in the Jujutsu integration.

I was getting the following error whenever I tried to review a commit (`tuicr -r @`) or selected a revision from the TUI:
```
Error: VCS command failed:

jj diff --from 41b940245b47b1d230b92b37c42ccddfcd351d99- \
        --to   41b940245b47b1d230b92b37c42ccddfcd351d99 \
        --git

Failed to parse revset: Syntax error

1 | ␛[38;5;4m41b940245b47b1d230b92b37c42ccddfcd351d99␛[39m-
```
After a bit of debugging I noticed the revision being passed to `jj` contains ANSI color escape sequences (ESC[38;5;4m...ESC[39m).

As a workaround, running:

`jj config set --repo ui.color never`

makes everything work again.

That makes me think tuicr is parsing colored output from `jj` and then reusing those revision strings when invoking `jj diff`.

I'm not familiar enough with the codebase to know if that's exactly what's happening, but it seems likely.

emrtnn:
This fixes #508  where the `jj` backend breaks when `ui.color` is enabled.

The root cause is that `tuicr` parses the output of `jj` programmatically, but by default `jj` may emit ANSI color sequences (`\x1b`). Those escape sequences can end up inside revision IDs and later get passed back to commands such as `jj diff`, resulting in revset parsing errors.

This change forces `jj` commands executed by the backend to use `--color=never`, ensuring all command output is machine-readable.

I've also added a regression test that verifies revision IDs do not contain ANSI escape sequences. The test fails without this change and passes with it.


agavra (approved):
nice find, thanks for the contribution @emrtnn 🙏

PR #511, opened 2026-07-29
Make long errors fully accessible

barlevalon:
## Summary

- keep status messages to one row and advertise `:messages` for errors
- show the complete current error in a full-screen, scrollable Messages view
- restore the originating diff or target-selector context on close
- promote PR-list failures to the shared error surface without dropping API response details
- handle keyboard and mouse scrolling, message replacement, terminal resizing, and Unicode display widths

`:messages` shows the current full error. Close it with `q` or `Esc`; navigation uses the existing help scrolling keys (`j/k`, arrows, `Ctrl-d/u`, `g/G`, Page Up/Down).

## Validation

- `cargo fmt --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test` — 1151 passed, 1 ignored
- `cargo build`
- `git diff --check`
- manually dogfooded at 70×16 with an oversized synthetic forge error, including full-screen rendering, selector return, and mouse/keyboard scrolling


agavra:
I've long wanted a feature like this, thanks @barlevalon 🙏 what do you think about making it a full screen thing instead of a popup/modal? it feels a bit distracting to have it that way. Then we can eventually add things like message number/timestamps/etc.. and make it a bit more of a log (which we should probably spool it to as well, but we can do that in the future)

barlevalon:
Thanks! I like that direction. It fits `:messages` better and leaves a cleaner path toward a proper message log. I’ll change the current single-message view to full-screen while keeping history, timestamps, and spooling out of scope for this PR. `q`/`Esc` will still return to the originating view.

barlevalon:
Updated in ef0c18e: `:messages` now opens a full-screen Messages view rather than a popup. The originating diff/selector view is restored on `q`/`Esc`; history, timestamps, and spooling remain intentionally out of scope.

rickgonzalez:
I chose this public PR to scan for my testing at Lenzon, I hope it's useful in some way:
https://www.lenzon.ai/viewer/cms84dh1r000dtexxsu8mhf13?voice=google-chirp3

PR #517, opened 2026-07-29
fix: Option for relative line numbers in the diff gutter

### LINKED ISSUE #509: Option for relative line numbers in the diff gutter
tuicr supports vim count prefixes (`{N}j`, `{N}k`), but the gutter shows absolute source
line numbers, so there is no way to read N off the screen. You have to count rows by eye.

Relative line numbers (`:set relativenumber`) are a standard Vim feature and exactly what
makes count-prefixed motions usable.

Proposal: a config option, e.g. `relative_line_numbers = false`, plus `:set relativenumber` /
`:set norelativenumber` / `:set relativenumber!` to toggle at runtime.

The numbers should be relative to rendered rows, not source lines, so they match what
`{N}j` actually does.

yhay81:
Fixes #509.

## Summary

Option for relative line numbers in the diff gutter

## Validation

- Mechanical gate: +162/-19, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ appearance = "system"        # or "dark" / "light"
 mouse = true
 leader = ";"                  # configurable prefix for leader shortcuts
 comment_vim = false           # vim modal editing in the review comment box
+relative_line_numbers = false # show rendered-row distances in the diff gutter
 review_watch_interval_ms = 1000 # set to 0 to disable persisted-review polling
 
 [[comment_types]]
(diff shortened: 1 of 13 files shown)


agavra (approved):
heh, this is super cool. I wasn't aware of this functionality in vim at all. a bit of a trip to see the numbers constantly change but really helpful. thanks for the path @yhay81

PR #523, opened 2026-07-30
Add github-dark-colorblind theme

Tacklebox:
Unified mode is very confusing with the default themes for red green colourblind people.


agavra (approved):
Thanks @Tacklebox!

PR #527, opened 2026-07-31
docs: document remaining config keys

CooperSheroy:
## Summary
- add `single_file_view` and `username` to the full config example
- document their defaults and behavior in the config options table

## Verification
- `git diff --check`
- direct inspection against `src/config/mod.rs` accepted keys

Note: `cargo test config::tests::should_parse_single_file_view_true config::tests::should_parse_username --lib` could not run locally because this cron environment has no installed Rust stable toolchain (`cargo` missing; `rustup run stable` reports stable-aarch64-apple-darwin is not installed).

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -37,6 +37,8 @@ transparent_background = true
 scroll_offset = 5
 no_update_check = false
 review_watch_interval_ms = 1000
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thanks for updating the docs!

Issue #438, opened 2026-06-26
File Tree shows files belonging to wrong directory 


Dynge:
Sometimes files show up as being in the test folder while they most certainly are not. I suspect it may be something with multi commit diffs.

agavra:
Ah hmm, might be the case if you move a file. I haven't seen this myself, if you have a repro I can look into it otherwise if you want to submit a patch fixing it on your workspace i'd be happy to review it.

Dynge:
This is not a case of files being moved. The image below shows an example.
The files in the box are under the folder `ChronoStream` however they appear to be under `BuildTests`.

I am using jj if that is relevant.

<img width="378" height="437" alt="Image" src="https://github.com/user-attachments/assets/129ab648-82cb-4ab0-b96c-55835af5c25b" />

Dynge:
I know a reproduction repository would be prefered. :D I thought I would show you a screen grab first in case that explains something. Maybe that the prefix of the folders are identical (`ChronoStream`)? :)

Dynge:
Reproduced with this test:


```
 #[test]
 fn test_interleaved_paths_stay_under_own_directory() {
     let mut h = TreeTestHarness::new(&[
         "ChronoStream/Subdir/file.cs",
         "ChronoStream.BuildTests/test.cs",
         "ChronoStream/root.cs",
     ]);
     h.expand_all();
 
     let items = h.build_visible_items();
     let root_file_idx = h
         .diff_files
         .iter()
         .position(|file| file.display_path() == Path::new("ChronoStream/root.cs"))
         .unwrap();
     let root_file_pos = items
         .iter()
         .position(|item| {
             matches!(item, FileTreeItem::File { file_idx, .. } if *file_idx == root_file_idx)
         })
         .unwrap();
 
     assert!(matches!(
         &items[root_file_pos],
         FileTreeItem::File { depth: 1, .. }
     ));
     assert!(matches!(
         items[..root_file_pos].iter().rev().find(|item| {
             matches!(item, FileTreeItem::Directory { depth: 0, .. })
         }),
         Some(
             FileTreeItem::Directory { path, .. }
         ) if path == "ChronoStream"
     ));
     assert!(matches!(
         &items[0],
         FileTreeItem::Directory { path, .. } if path == "ChronoStream"
     ));
 }
```

agavra:
Thanks! thaths really helpful. I'll take a look.

PR #528, opened 2026-07-31
fix(tree): keep a directory's files together when a sibling shares its prefix

Dhevenddra:
## Problem

Reported in #438: files render under a directory they don't belong to. @Dynge's screenshot shows files from `ChronoStream/` appearing nested under `ChronoStream.BuildTests/`.

`sort_files_by_directory` groups files into a `BTreeMap` keyed by the joined parent path string. `.` is ASCII 46 and `/` is 47, so `ChronoStream.BuildTests` sorts between `ChronoStream` and `ChronoStream/Subdir`. That splits the `ChronoStream` subtree into two runs that are no longer adjacent.

`build_visible_items` emits a directory header only the first time it sees a directory, and it derives each file's indent from its ancestor count. Once the subtree is split, that function suppresses the second run's header, and those files render indented under whichever directory landed between the halves.

Any sibling whose name continues with a character below `/` hits this. `.` and `-` are the common ones, so `foo.Tests` next to `foo/`, or `my-lib` next to `my/`.

## Fix

Key the map on the parent's path components rather than the joined string. Comparing component by component keeps a parent adjacent to its own children.

## Verification

Added `test_interleaved_paths_stay_under_own_directory` using @Dynge's repro paths. It drives a real `App` through `sort_files_by_directory` and `build_visible_items`. I deliberately did not use the local `TreeTestHarness`: it keeps its own copy of the tree logic, so a test written against it passes no matter what the production path does. I checked that first, and a harness-based version of this test did pass against the unfixed code.

With the fix reverted and the test left alone, it fails with the reported symptom:

```
left:  [("ChronoStream/", 0), ("ChronoStream/root.cs", 1), ("ChronoStream.BuildTests/", 0), ("ChronoStream.BuildTests/test.cs", 1), ("ChronoStream/Subdir/", 1), ("ChronoStream/Subdir/file.cs", 2)]
right: [("ChronoStream/", 0), ("ChronoStream/root.cs", 1), ("ChronoStream/Subdir/", 1), ("ChronoStream/Subdir/file.cs", 2), ("ChronoStream.BuildTests/", 0), ("ChronoStream.BuildTests/test.cs", 1)]
```

`cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` are both clean. `cargo test` gives 1196 passed, 2 failed. The two failures are `vcs::git::libgit2::tests::should_discover_worktree_with_relativeworktrees_extension` and `vcs::git::tests::default_preference_routes_reftable_repo_to_cli`, which fail identically on an unmodified checkout here. My git is 2.43.0, which predates both extensions.

### DIFF
--- a/src/app/tests/tree_tests.rs
+++ b/src/app/tests/tree_tests.rs
@@ -1,5 +1,6 @@
 use crate::app::*;
-use crate::model::{DiffFile, FileStatus};
+use crate::model::{DiffFile, DiffLine, FileStatus};
+use crate::vcs::traits::{VcsBackend, VcsInfo, VcsType};
 
 fn make_file(path: &str) -> DiffFile {
     DiffFile {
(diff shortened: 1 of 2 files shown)


agavra (approved):
this makes sense to me, thanks!

Issue #481, opened 2026-07-18
always show commit


agavra:
right now we only display the commit when reviewing a specific revision (using `-r <sha>`), I find it useful to always display it

PR #529, opened 2026-07-31
feat(ui): always show the commit under review in the header

Dhevenddra:
Relates to #481.

## What changed

`header_source_chunk` returned `None` for a plain working-tree review, and a bare label for the staged and unstaged variants. All four diff against HEAD, so the commit being reviewed existed but was never displayed. Only `-r <sha>` named a commit.

| review mode | before | after |
|---|---|---|
| working tree | `git:main` | `git:main · commit abc1234` |
| staged | `git:main · staged` | `git:main · staged · commit abc1234` |
| unstaged | `git:main · unstaged` | `git:main · unstaged · commit abc1234` |
| staged + unstaged | `git:main · staged + unstaged` | `git:main · staged + unstaged · commit abc1234` |
| `-r <sha>` | `git:main · commit fedcba9` | unchanged |
| PR mode | unchanged | unchanged |
| pristine | unchanged | unchanged |

The commit-range arms are untouched, so a revision review still names only the revision it is reviewing and nothing renders the sha twice.

## Two cases deliberately excluded

**Pristine mode** keeps its existing `PRISTINE · <head> · N files` chip. Its `head_commit` holds a synthetic `pristine:<head_or_none>:<hash>` key rather than a sha, so slicing it would print `pristin`.

**A repository with no HEAD** falls back to the previous labels rather than showing an empty chip, so a fresh `git init` reads the same as before.

## Tests

Five cases added to the header snapshot module:

- working-tree review reports `commit abcdef0`
- `staged` and `staged + unstaged` suffix the same chip
- an empty `head_commit` yields `None` for working tree and a bare `staged` otherwise
- a `-r <sha>` review still reports only its own revision
- the string actually reaches the rendered header buffer, not just the chunk builder

Checked the tests catch this: with the four match arms reverted and the tests kept, three of the five fail. The other two guard behaviour that must not change, so they pass either way by design.

I also renamed the enclosing test module from `pr_header_snapshot_tests` to `header_snapshot_tests`, since it now holds local-review cases too. Happy to drop that if you would rather keep the old name.

`cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` are clean. `cargo test` is 1200 passed, 2 failed; both failures are `vcs::git` `reftable` and `relativeworktrees` cases that fail identically on an unmodified checkout here, since my git is 2.43.0 and predates those extensions.

### DIFF
--- a/src/ui/status_bar.rs
+++ b/src/ui/status_bar.rs
@@ -158,14 +158,39 @@ pub fn render_header(frame: &mut Frame, app: &App, area: Rect) {
     );
 }
 
-/// Short, lowercase description of the active review source. Returns `None`
-/// for plain working-tree review (no extra label needed beyond `vcs:branch`).
+/// Short form of the HEAD sha, or `None` when there is no real commit to name.
+///
+/// Pristine sessions store a synthetic `pristine:<head>:<hash>` key rather than
+/// a sha, and that mode already renders its own chip further down the header.
+/// An empty repository has no HEAD at all.
+fn head_commit_label(app: &App) -> Option<String> {
+    if app.is_pristine_mode {
+        return None;
+    }
+    let head = app.vcs_info.head_commit.as_str();
+    (!head.is_empty()).then(|| head[..7.min(head.len())].to_string())
+}
+
+/// Suffix a working-tree label with the commit being diffed against.
+fn with_head_commit(label: &str, app: &App) -> String {
+    match head_commit_label(app) {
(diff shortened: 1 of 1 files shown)


agavra:
The original issue was a bit unclear, the intention was to show the commit message. We need a bit of design for this so that it will be clear how it works when there are multiple commits being reviewed. 

Otherwise this makes sense to me, but i will not close the ticket.

agavra (approved):

PR #534, opened 2026-08-01
fix(diff): keep hunk lines whose content starts with `---` or `+++`

thejesh23:
## The bug

`parse_hunk` skips any hunk body line starting with `---` / `+++`, intending to ignore stray file-header lines:

https://github.com/agavra/tuicr/blob/246f839e83ac4149489c43357e5ee2230bd16676/src/vcs/diff_parser.rs#L295-L310

The check runs against the **prefixed patch line**, not the line's content. A deleted line is emitted as `-` + content, so:

- deleting `---` (YAML document separator / Markdown front-matter fence) emits `----` → matches → dropped
- deleting `-- count the rows` (SQL/Lua/Haskell/Ada/Elm comment) emits `--- count the rows` → matches → dropped
- adding `++i;` emits `+++i;` → matches → dropped

Worse, `continue` runs **before** `old_lineno` / `new_lineno` is incremented, so the dropped line's number is reused by the next line — every following line number on that side of the hunk shifts by one, compounding per dropped line.

## Repro

```
diff --git a/post.md b/post.md
--- a/post.md
+++ b/post.md
@@ -1,5 +1,2 @@
----
-title: hello
----
 intro
-old body
+new body
```

Parsed today — 4 lines instead of 6, both `---` deletions invisible, old-side numbers wrong:

```
Deletion old=Some(1) "title: hello"   <- actually old line 2
Context  old=Some(2) "intro"          <- actually old line 4
Deletion old=Some(3) "old body"       <- actually old line 5
Addition new=Some(2) "new body"
```

## Impact

The reviewer never sees that the line was deleted. And because `find_line_with_counterpart` in `src/forge/submit.rs` uses `DiffLine::old_lineno` / `new_lineno` verbatim as the forge comment position, a comment placed later in that hunk is **posted on the wrong upstream line**.

Reachable on the GitHub/GitLab PR path (`forge/pr_open.rs`, `app/pr.rs` both call `parse_unified_diff`), and on the jj, hg and sparse-Git CLI backends. The default libgit2 backend is unaffected.

## The fix

Delete both guards. They are dead for their stated purpose: `parse_unified_diff_lines` only starts a file on a `diff --git ` / `diff ` header, `parse_file_header` consumes the `---`/`+++` pair before the first `@@`, and the hunk loop already breaks on the next `diff ` / `@@` — so a genuine file header can never reach that code. A comment in their place records why no check belongs there.

## Verification

Three regression tests added (deleted YAML fence, deleted SQL comment, added `++i;`), each asserting both the retained line and the unshifted numbers that follow. Confirmed they fail on `main` before the fix:

```
should_keep_deleted_lines_whose_content_starts_with_dashes  left: 4  right: 6
should_keep_deleted_sql_comment_lines                       left: 1  right: 2
should_keep_added_lines_whose_content_starts_with_plus_signs left: 2 right: 3
```

`cargo test` → **1203 passed, 0 failed**. `cargo fmt --all --check` and `cargo clippy -- -D warnings` clean.

### DIFF
--- a/src/vcs/diff_parser.rs
+++ b/src/vcs/diff_parser.rs
@@ -292,19 +292,19 @@ where
             continue;
         }
 
+        // No `---`/`+++` header check here: a hunk body line is never a file
+        // header. The outer loop only starts a file on a `diff --git `/`diff `
+        // line, `parse_file_header` consumes the `---`/`+++` pair before the
+        // first `@@`, and this loop breaks on the next `diff `/`@@` above. A
+        // check here would instead match *content*: deleting a line beginning
+        // with `--` (a SQL/Lua/Haskell comment, a YAML `---` separator, a
+        // Markdown front-matter fence) emits `---…`, and skipping it would drop
+        // the line and desynchronize every following line number in the hunk.
         let (origin, content, old_ln, new_ln) = if let Some(stripped) = line.strip_prefix('+') {
-            if line.starts_with("+++") {
-                // Skip +++ header lines
-                continue;
-            }
             let ln = new_lineno;
             new_lineno += 1;
             (LineOrigin::Addition, stripped, None, Some(ln))
         } else if let Some(stripped) = line.strip_prefix('-') {
-            if line.starts_with("---") {
-                // Skip --- header lines
-                continue;
-            }
             let ln = old_lineno;
             old_lineno += 1;
             (LineOrigin::Deletion, stripped, Some(ln), None)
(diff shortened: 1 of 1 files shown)


thejesh23:
CI note: the only red check is `Update command (windows-latest)`, which fails at `Error: failed to fetch update metadata: http status: 403` in the live self-update step — a GitHub API rate-limit response, not a code failure. It is unrelated to this change (which touches only `src/vcs/diff_parser.rs`): the same job is failing on `main` at `246f839` and on the two commits before it, and it passed on my other PR (#535) minutes later on the same base. `Check`, `Format`, `Clippy` and `Test` are all green here. I do not have rights to re-run the job; a maintainer re-run should clear it.

agavra (approved):
Thanks for the fix!

PR #537, opened 2026-08-01
fix(release): build static musl binaries for Linux artifacts

### LINKED ISSUE #503: The binaries are not static
I tried installing this via the `curl install.sh | sh` and by directly downloading the binaries, but in both cases I hit this snag.

```
$ file tuicr
tuicr: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=c59276cf98afd143c5ab0e742ec0fbf458ee93fe, not stripped

$ ./tuicr
./tuicr: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./tuicr)
```

Would you be open to a PR statically linking the built binary with musl or otherwise providing alternatives for when glibc 2.39 is not available?

mvanhorn:
## Summary

Builds the Linux release artifacts against musl instead of glibc, so the published binaries are statically linked and run on distros older than the builder. Closes #503.

The Linux jobs built with `cargo build --release` on `ubuntu-latest`, producing binaries dynamically linked against that runner's glibc. Anyone on an older distro got a glibc version error rather than a working binary.

Changes:

- `.github/workflows/release.yml`: each matrix entry gains a `build_target`; Linux maps to the musl triple while macOS and Windows keep their existing target. Installs `musl-tools`, points the musl linker and `CC` at `musl-gcc`, builds with `--target`, and adds a verification step that runs `--version` on the produced binary and fails if `readelf` still shows an `INTERP` segment. The published artifact names are unchanged, so download URLs stay stable.
- `Cargo.toml`: `git2` moves to target-conditional dependencies. Non-musl keeps the current default-features-off build; musl adds `zlib-ng-compat`, because the vendored zlib otherwise links against the host library and defeats the static build.
- `Cargo.lock`: picks up `zlib-ng-compat`'s tree.
- `.github/workflows/ci.yml`: a musl build job so a change that breaks the static build fails in CI rather than at release time.
- `RELEASE.md`: note that Linux artifacts are static musl builds.

## Validation

From a clean checkout of this branch:

```text
cargo build           -> ok
cargo fmt --check     -> clean
cargo metadata        -> ok (target-conditional deps resolve)
```

I could not exercise the musl path locally; that needs a Linux runner with `musl-tools`, and I am on macOS. The parts I could not run are the ones the new CI job and the `readelf` INTERP check now cover, which is deliberate: the verification step is written to fail loudly on the runner rather than silently ship a dynamic binary again.

Worth your eye before merge:

1. `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` are kept as the artifact `target` names while building musl underneath. That preserves existing download URLs, but the filename now says gnu for a musl binary. If you would rather rename the artifacts and accept the URL break, that is a one-line change per matrix entry.
2. The `libssl-dev`/`pkg-config` install is replaced by `musl-tools`. Nothing in the current dependency tree needs OpenSSL (`git2` is default-features-off), so this should be safe, but you know the tree better than I do.

AI was used for assistance.


RELEASE.md
```diff
@@ -16,8 +16,8 @@ This project uses an automated release workflow via GitHub Actions.
    - Publishes to crates.io
    - Creates GitHub Release with release notes
    - Builds and uploads binaries for:
-     - `x86_64-unknown-linux-gnu` (Linux x64)
-     - `aarch64-unknown-linux-gnu` (Linux ARM64)
+     - `x86_64-unknown-linux-gnu` (Linux x64, static musl payload)
```
wcampbell0x2a: I really don't think lying about the libc is a good idea.

mvanhorn:
Fair point, dropped the gnu names in 4339ba2.

The matrix had a separate target field that existed only so the archives could keep their old -unknown-linux-gnu names while building against musl. That's gone; the artifact name now comes from build_target, so the Linux archives are x86_64-unknown-linux-musl and aarch64-unknown-linux-musl.

That does break the existing Linux download URLs, which is what the old naming was protecting. RELEASE.md is updated to match, including the checksum commands for the Homebrew formula. Flagging it in case the install script or tuicr update pin those URLs anywhere I haven't found.

lllamnyp (approved):
Thanks for picking up #503, this looks great!

lllamnyp:
Thanks for taking #503.

I was testing out the build and found that running `tuicr update` replaces the statically linked binary with a dynamically linked one. Here is where this appears to be pinned:

https://github.com/agavra/tuicr/blob/4339ba28b7b737013ce4546d88c9aae16bb7b829/src/update/install/source.rs#L28-L29

Two things that might bite once this lands, though you'd know better than me: after a release that ships no gnu Linux archive, `tuicr update` on existing installs looks like it would fail outright rather than upgrade, since the asset is looked up by exact name. And the triple is baked in at compile time, so a fix landing in that same release wouldn't reach the binaries already installed.

mvanhorn:
Good catch on both counts, thanks for actually running the built binary.

The release workflow now builds `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` alongside the musl targets, with `-C target-feature=+crt-static` so the gnu archives are statically linked too. `musl-tools` and the musl linker config are gated on the `-musl` suffix so the gnu jobs skip them.

That keeps the exact asset names existing installs look for, so `tuicr update` on a gnu install still resolves and now lands a static binary rather than swapping in a dynamic one.

The compile-time triple point stands: anything already installed keeps whatever triple it was built with, so this only takes effect from the next release onward. Nothing to do about already-shipped binaries short of a re-install, which seems acceptable given the alternative was update breaking outright.

agavra:
We now rely on the official hombrew tap, I'm not sure if we need to change anything to get it to update after that. How are you intending to install the static binaries? Otherwise LGTM, I'm happy to merge it anyway.

agavra:
CI failure is unrelated, rebasing on main should fix this.

lllamnyp:
> How are you intending to install the static binaries?

Seeing as I brought up the original issue, I'll answer this. I **usually** prefer to install them by directly downloading them from GitHub's releases. The pipe-curl-to-shell approach also appears to be doing more or less the same thing. I don't think much if anything would change, regardless of install method.

mvanhorn:
Rebased onto main at 08aa4e9, so the unrelated CI failure should be gone. cargo check is clean and both workflow files still parse.

@lllamnyp thanks, that settles it. Direct download from the releases page is what the static musl artifacts are built for, and the curl-to-shell path pulls the same archives, so neither install method needs anything extra here.

PR #559, opened 2026-08-05
diff view: hide the Review Comments section while it is empty

thiblahute:
The `═══ Review Comments ═══` section always rendered, even before any
review comment existed. Hide the header until the section has content — a
remote review summary, a local review comment, or a visible review-level
remote thread. Per review feedback this is now simply the behavior — no
config option.

The gate goes through a shared `App::show_review_comments_header()` used
by the unified and side-by-side renderers, the annotation builder, and the
scroll height model (`review_comments_section_height`), so the display,
the cursor/click model, and scrolling stay in sync.

Covered by a unit test (`review_comments_header_hidden_while_empty`);
`cargo fmt`, `check`, `clippy -D warnings`, and `cargo test` all pass
locally.

### DIFF
--- a/src/app/annotations.rs
+++ b/src/app/annotations.rs
@@ -85,10 +85,11 @@ impl App {
             }
         }
 
-        // The review-comments header is omitted in single-file view (see
-        // the matching guard in `src/ui/diff_unified.rs`), so the
-        // annotation list mirrors the render.
-        if !self.is_single_file_view {
+        // The review-comments header is omitted in single-file view and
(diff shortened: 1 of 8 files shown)


agavra:
I'm happy making this the default with no configuraiton for it, no need to add an extra config when I think everyone would want this!

thiblahute:
Indeed, makes sense, done :-)

thiblahute:
This one should be ready :-)

agavra (approved):
Thanks @thiblahute :)

PR #561, opened 2026-08-05
fix: reindex recovered PR sessions

### LINKED ISSUE #449: Re-index PR sessions recovered from deterministic fallback path
## Bug

When a PR session is recovered from its deterministic JSON file path, the TUI can attach to the correct session without repairing `index.json`.

The problematic flow is:

1. A PR session file exists for the current `PrSessionKey` / head SHA.
2. The manifest has no current-head PR entry, or has a stale entry for the PR slug.
3. The PR open path recovers the session from the deterministic file path.
4. `reset_persisted_session_tracking()` sees the path already exists and only marks it active in `active_sessions.json`.
5. Nothing upserts the recovered session back into `index.json` until a later save/comment.

## Impact

The TUI uses the correct recovered session, but `tuicr review list` and `tuicr review add <slug>` still resolve through the stale or missing manifest entry.

This is independent of the hunkless-diff carry-forward bug.

## Expected behavior

When the exact-file fallback successfully loads a PR session, storage should also repair the manifest entry for that PR slug/head before the session is considered active.

## Possible fix

Add a storage helper that re-indexes an existing session file, or route the fallback through a save/reindex path that upserts the manifest without changing session contents.

Regression test: create a current-head PR session JSON at the deterministic path with a missing/stale manifest, recover it, then assert `ReviewStore::resolve_pr_session("gh:owner/repo/pr/N")` returns the recovered path.

Guflly:
Re-index PR sessions recovered from their deterministic path so the review list and review add see them immediately. The repair only updates the manifest and leaves the session JSON unchanged.

Adds regression coverage for missing and stale manifest entries.

Tests:
- cargo test should_reindex_recovered_pr_session
- cargo fmt --all -- --check

Closes #449

### DIFF
--- a/src/app/tests/target_selector_tests.rs
+++ b/src/app/tests/target_selector_tests.rs
@@ -1189,7 +1189,7 @@ fn should_keep_saved_pr_session_through_quit_reopen_and_same_head_reload() {
 }
 
 #[test]
-fn should_use_persisted_new_head_session_instead_of_carrying_old_head_state() {
+fn should_use_and_reindex_persisted_new_head_session() {
     // given an old-head PR session with reviewed state
     let _reviews = TestReviewsDir::new();
     let mut app = build_app();
@@ -1229,7 +1229,10 @@ fn should_use_persisted_new_head_session_instead_of_carrying_old_head_state() {
             CommentType::from_id("note"),
             None,
         ));
+    let persisted_path = crate::persistence::storage::session_path(&persisted_b).unwrap();
+    let slug = crate::slug::Slug::from(persisted_b.pr_session_key.as_ref().unwrap()).to_string();
     write_session_file_without_manifest(&persisted_b);
+    let persisted_contents = std::fs::read(&persisted_path).unwrap();
 
     // when the PR reload advances to that head
     let backend_b = Box::new(FakeForgeBackend::open_pr_details(
@@ -1248,6 +1251,12 @@ fn should_use_persisted_new_head_session_instead_of_carrying_old_head_state() {
     let changed_review = app.session.files.get(&changed_path).unwrap();
     assert_eq!(changed_review.file_comments.len(), 1);
     assert_eq!(changed_review.file_comments[0].content, "new-head draft");
+    let resolved = crate::review_store::ReviewStore::new()
+        .resolve_pr_session(&slug)
+        .unwrap()
+        .unwrap();
+    assert_eq!(resolved.path(), persisted_path);
+    assert_eq!(std::fs::read(&persisted_path).unwrap(), persisted_contents);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


src/app/tests/target_selector_tests.rs
```diff
@@ -1189,7 +1189,7 @@ fn should_keep_saved_pr_session_through_quit_reopen_and_same_head_reload() {
 }
 
 #[test]
-fn should_use_persisted_new_head_session_instead_of_carrying_old_head_state() {
+fn should_use_and_reindex_persisted_new_head_session() {
```
benja2998: This function name is really long

src/app/tests/target_selector_tests.rs
```diff
@@ -1229,7 +1229,10 @@ fn should_use_persisted_new_head_session_instead_of_carrying_old_head_state() {
             CommentType::from_id("note"),
             None,
         ));
+    let persisted_path = crate::persistence::storage::session_path(&persisted_b).unwrap();
+    let slug = crate::slug::Slug::from(persisted_b.pr_session_key.as_ref().unwrap()).to_string();
```
benja2998: Can't these be multiple lines for readability?

src/persistence/storage.rs
```diff
@@ -342,13 +359,22 @@ fn save_session_in_dir_unlocked(session: &ReviewSession, reviews_dir: &Path) ->
...
+
+fn upsert_session_manifest(
+    session: &ReviewSession,
+    reviews_dir: &Path,
+    slug: &Slug,
+    relative: PathBuf,
+) -> Result<()> {
     let mut manifest = manifest::load_manifest(reviews_dir).unwrap_or_default();
-    let anchor = manifest_anchor_for(&slug);
+    let anchor = manifest_anchor_for(slug);
```
benja2998, line 374: What exactly does this do?

benja2998 (changes requested):

src/app/tests/target_selector_tests.rs
Guflly: Renamed it to should_reindex_recovered_pr_session.

src/app/tests/target_selector_tests.rs
Guflly: Split the session-key lookup into a separate variable.

src/persistence/storage.rs
Guflly: It rebuilds the manifest entry for the existing session file. That lets slug-based lookups find the recovered session without rewriting its JSON; normal saves use the same helper.

Guflly:
Rebased on current main. Cargo fmt, cargo check, and the recovered-session regression test pass.

agavra (approved):
Thanks @Guflly :)

PR #563, opened 2026-08-05
add basic emacs keybindings

### LINKED ISSUE #499: [Feature Request] Support for basic emacs keybindings
Emacs keybindings are the standard keybindings. They're used by default by standard software like Readline (which is used by Python's REPL, Bash, ETC). Most shells, like Fish and Zsh, also include Emacs keybindings as the defaults. OS X allows using a subset of Emacs keybindings in text input areas.
It would make sense to have at least a toggle for Emacs keybindings, while keeping Vim keys the default as they seem to be a selling point for your software.

benja2998:
this pull request adds basic emacs keybindings while preserving vim keybindings. It makes sense to merge because Emacs keybindings are the standard keybindings in CLI software. Closes #499


benja2998:
Closing because of https://github.com/agavra/tuicr/issues/499#issuecomment-5197657019

PR #566, opened 2026-08-06
feat: add diff_watch_interval + fix some bugs

### LINKED ISSUE #488: Feature Request - diff_watch_interval_ms for local diff changes
Love tuicr, use it every day.

For my workflow, I review local diffs. I keep having to `:e` to refresh and see the updated diff.

Just like we have `review_watch_interval_ms` for  watching for review comments, would you be open to adding `diff_watch_interval_ms` to auto refresh the diff?

I'm willing to contribute this myself. I just want to make sure that aligns with the vision for the tool before putting in the effort.

arjansingh:
# Description

1. `diff_watch_interval_ms`: You can set an interval in `config.toml` for `tuicr` to watch for changes. Each tick runs a cheap check first, and only spawns a thread to fetch the real diff when something actually changed. That keeps an idle ticks fast. It's off by default and `0` also disables it, so nothing changes unless you opt in. It disabled for PR reviews and `--all-files`. I confirmed this works for new commits, staged changes, and unstaged changes.

2. fix nav scrolling: I noticed that when I `j` toward the bottom of a long diff, sometimes it would stop short and leave the last lines unreachable. It only happens when lines wrap. I fixed that. Of note, `G` has the same symptom through a different function, `jump_to_bottom`, which I have not touched in the interest of finishing #488. Should be a similar fix.

3. fix review marking in a commit: I noticed that pressing `r` or `R` on a file that only exists in a commit did nothing at all. Commenting on that same file fails too, with `session does not contain file`. The PR side already registers these files correctly, so the local side was just inconsistent. Not good if you or your agent are reviewing a huge set of commits and want to mark something off or point out something that happened in the past. I fixed that too.

Tests added for features and fixes of course. They are actually the majority of this diff.

Closes #488.


arjansingh:
I didn't add it to this PR, but I think there are also some ways we can improve rendering and refreshing of really large diffs. (200+ files).

This watcher works pretty well with ~100-ish file diffs right now.

Files appear to be the limiting perf factor more than size of diff.

src/ui/status_bar.rs
```diff
@@ -16,6 +16,8 @@ use crate::ui::styles;
 /// Maximum visible completion candidates in the command prompt popup.
 const COMMAND_COMPLETION_MAX_ROWS: usize = 7;
 
+const DIFF_WATCH_GLYPH: &str = "\u{25c9}"; // ◉
```
agavra: I don't think we should add this to the footer, it's not particularly helpful to permanently have hat there

agavra (approved):
Thanks @arjansingh this is cool. When you said that it works for "new commits" what do you mean by that?

Otherwise just one small nit and I think we need to rebase/resolve a minor merge conflict and we're good to go.

src/ui/status_bar.rs
arjansingh: sounds good. i can remove.

arjansingh:
> When you said that it works for "new commits" what do you mean by that?

I meant that when I commit changes, the commit list updates with the "new commit" I just added.

src/ui/status_bar.rs
arjansingh: Removed.

agavra:
Thanks @arjansingh - LGTM!

PR #570, opened 2026-08-06
feat(ui): copy the comment at the cursor with Y

### LINKED ISSUE #569: Copy a single comment to the clipboard (Y at cursor, symmetric with dd)
`dd` deletes the comment at the cursor, but there's no way to copy just that one comment. `y` in normal mode exports the whole review, and `y` in visual mode copies diff text only — `content_for_side` (`src/app/navigation.rs:540`) returns `None` for comment annotations, so dragging over a comment box yanks nothing.

The workaround is dropping to the CLI:

```bash
tuicr review comments --session "<slug>" | jq -r '.[] | select(.id == "…") | .content' | pbcopy
```

which means leaving the TUI to find the id of a comment I'm already looking at.

Use case: pasting one comment into Slack or into an agent prompt, without the other 8 comments in the session.

Proposal: `Y` on a comment row copies that comment's content. Symmetric with `dd`, and the "comment at cursor" plumbing already exists from #443.

Dhevenddra:
Fixes #569.

## Problem

`dd` deletes the comment at the cursor, but there was no way to copy just that one. `y` exports the whole review, and visual-mode `y` copies diff text only: `content_for_side` (`src/app/navigation.rs:540`) returns `None` for comment annotations, so dragging over a comment box yanks nothing. Getting one comment out meant leaving the TUI for `tuicr review comments --session ... | jq`, which needs an id you cannot see from the box you are looking at.

## Fix

`Y` in normal mode copies the content of the comment under the cursor.

- `App::comment_content_at_cursor` resolves through `find_comment_at_cursor`, the same lookup `dd` and `i` use, so `Y` and `dd` always agree on which comment the cursor is on. It carries over the side guard the delete path uses for line comments.
- The handler copies with `copy_text_to_clipboard`, so pbcopy, OSC 52, xclip and wl-copy behave exactly as they do for `y`, including the "(via terminal)" message.
- `y` and visual-mode `y` are untouched.

Remote threads are deliberately out of scope: they are read-only in v1 and their bodies live in `forge_review_threads` rather than the session. When the cursor is on one, the message names the forge instead of claiming there is no comment.

Help popup and `docs/KEYBINDINGS.md` both list the new key.

## Evidence

`cargo test`:

```
test result: FAILED. 1300 passed; 2 failed; 2 ignored
```

The two failures are `should_discover_worktree_with_relativeworktrees_extension` and `default_preference_routes_reftable_repo_to_cli`. Both fail identically on an unmodified checkout of `8323f13` here, because this box runs git 2.43.0, which predates the `relativeworktrees` and `reftable` extensions. Nothing else fails.

New tests:

- `should_yank_only_the_comment_under_the_cursor` puts two comments on different lines and checks each one yanks its own content, not its neighbour's.
- `should_yank_nothing_when_cursor_is_not_on_a_comment` pins the `None` path on a plain diff line.
- `should_map_lowercase_and_uppercase_y_to_separate_yank_actions` pins `y` and `Y` to different actions.

`cargo fmt --all --check` and `cargo clippy -- -D warnings` are both clean. (`clippy --all-targets` reports a pre-existing `items_after_test_module` on `src/handler.rs`, unrelated to this change.)

One note on process: I could not dogfood this through the TUI itself, since I am on a Windows box driving the build through WSL2 without an interactive terminal for tuicr. The diff is small and self-reviewed, but I have not watched `Y` copy a comment in a live session, so a quick manual check on your side would be worth it before merge.

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -121,6 +121,7 @@ Shown below the file tree when local comments or visible remote PR threads exist
 | `A` | Edit comment at cursor with text cursor at end (vim mode only) |
 | `e` | Open focused file in `$EDITOR` |
 | `y` | Copy review to clipboard |
+| `Y` | Copy the comment at cursor to clipboard |
 
 ## Visual mode
 
--- a/src/app/comments.rs
+++ b/src/app/comments.rs
@@ -419,6 +419,41 @@ impl App {
         }
     }
 
+    /// Content of the comment at the current cursor position, if any.
+    /// Resolves through the same lookup `dd` and `i` use, so `Y` yanks
+    /// exactly the comment the cursor is sitting on.
+    pub fn comment_content_at_cursor(&self) -> Option<String> {
+        match self.find_comment_at_cursor()? {
+            CommentLocation::Review { index } => self
+                .session
+                .review_comments
+                .get(index)
+                .map(|c| c.content.clone()),
+            CommentLocation::File { path, index } => self
+                .session
+                .files
+                .get(&path)
+                .and_then(|review| review.file_comments.get(index))
+                .map(|c| c.content.clone()),
+            CommentLocation::Line {
+                path,
+                line,
+                side,
+                index,
+            } => self
+                .session
+                .files
+                .get(&path)
+                .and_then(|review| review.line_comments.get(&line))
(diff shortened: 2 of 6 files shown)


agavra (approved):
LGTM! Thanks @Dhevenddra

PR #571, opened 2026-08-07
feat(search): highlight search matches in the diff

### LINKED ISSUE #520: Highlight search matches and provide a way to clear them
It would be helpful if tuicr highlighted text matching the current search query. This would make it easier to spot matches while navigating through a review.

Ideally, the highlights would remain visible after leaving search mode. There should also be a way to clear them when they are no longer needed, similar to Vim’s :noh command.

I’m not opinionated about the exact keybinding or command. My main goal is to make search results visually persistent and still providing an easy way to disable the highlighting.

I am willing to create the PR for this if this is something the project wants.

Dylan-MSML:
Closes #520

Here I added the vim style search feedback for `/`:

1. The matching lines are highlighted in unified and side-by-side views, 
2. the status bar shows a `[current/total]` counter
3. `n` / `N` cycle through the results, wrapping past the last match back to the first. The tool did not do this before so I understand if this should to do this in a seperate PR. But in the issue I created it was kind of implied that this was wanted.
4. The enabling/disabling of highlighting is also configurable

I decided to NOT add the `:noh`  command as when playing around with it it felt unneeded. `Esc` does the job well enough and is way more intuitive.

I also made sure that when we resume searches we also turn on the highlighting.

short summary of how I implemented this:
When we search we `fold` the pattern once and store the matching line indices in a sorted list. This gets recomputed whenever the diff structure changes. Renderers ask one accessor per line, which binary-searches that list and only then we run the highlighter to find the exact character ranges to paint. 
The matcher and highlighter share the same case-folding, so they always agree. `n/N` and the [current/total] counter reuse the same list, and the highlight color is blended from existing theme colors (I did consider adding another color token but I don't know if this is something the projects wants to add)

Screenshots:
image


agavra (approved):
This is really helpful, thanks @Dylan-MSML! I think next up we probably want `/` to use regex instead of exact matches, but that's a separate issue

PR #574, opened 2026-08-07
fix: prevent cursor-overflow panic on PR reload that shrinks the diff

### LINKED ISSUE #573: panic when reloading a PR of shorter length
I was using tuicr local built from this branch. I loaded up a PR while I had the main branch checked out. In another bash window, I checked out the PRs source branch. Made a change. Pushed the change remote. When I swithched back to the other bash window running tuicr, I issued command :e to refresh the pr, and tuicr panic crashed w error; 

[11:00]  Bash main ≡                                                                                                                                                            1s 391ms
   ~\source\repos\Comp.Submissions
  !❯ ../tuicr/target/debug/tuicr.exe
  tuicr-session: _git/Comp.Submissions@main/worktree/06ca818

  thread 'main' (13296) panicked at src\app\navigation.rs:54:32:
  attempt to subtract with overflow
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

joshvito:
fixes #573

A PR `:e` reload can replace the diff with a shorter one while leaving `cursor_line` parked past the new end (the same-head reload branch, and a restored overview cursor captured from the taller old diff, never reclamp it). The next `cursor_down` then clamps the cursor *up* to the new `max_cursor_line` — below `prev_cursor` — underflowing the `cursor_line - prev_cursor` usize subtraction and panicking with "attempt to subtract with overflow" at navigation.rs.

- cursor_down: use `saturating_sub` for the cursor-movement delta, matching the existing `cursor_up` siblings.
- finish_pr_reload / reload_pull_request_with_backend: clamp `cursor_line` into the reloaded diff's bounds after the swap/restore.
- add a regression test that reproduces the stale-cursor overflow.

### DIFF
--- a/src/app/tests/scroll_behavior_tests.rs
+++ b/src/app/tests/scroll_behavior_tests.rs
@@ -725,3 +725,27 @@ fn move_cursor_to_annotation_wrap_on_scrolls_target_fully_into_view() {
         "already-visible target should not change scroll_offset"
     );
 }
+
+#[test]
+fn cursor_down_with_stale_cursor_past_end_does_not_underflow() {
+    // Regression: a PR reload can shrink the diff and leave `cursor_line`
+    // parked past the new `max_cursor_line`. The next `cursor_down` clamps
+    // the cursor *up* to the new max (below `prev_cursor`), which used to
+    // underflow the `cursor_line - prev_cursor` usize subtraction and panic
+    // (`attempt to subtract with overflow` at navigation.rs).
+    let mut app = build_scroll_app(10, 20, 0);
+    // Overview (multi-file) path exercises the `cursor_line - prev_cursor`
+    // subtraction; the single-file walk branch clamps separately.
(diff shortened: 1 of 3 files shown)


joshvito:
I want to test this before I publish.

src/app/tests/scroll_behavior_tests.rs
```diff
@@ -725,3 +725,27 @@ fn move_cursor_to_annotation_wrap_on_scrolls_target_fully_into_view() {
         "already-visible target should not change scroll_offset"
     );
 }
+
+#[test]
+fn cursor_down_with_stale_cursor_past_end_does_not_underflow() {
+    // Regression: a PR reload can shrink the diff and leave `cursor_line`
```
joshvito: just testing that this change didn't break basic commenting functionality;

agavra (approved):
nice catch @joshvito 🙏

PR #581, opened 2026-08-09
fix: keep the TUI on screen when opening a windowed editor

### LINKED ISSUE #580: Opening file in windowed editor causes flicker
Opening a file in a windowed/external editor like VSCode causes a flicker because the TUI gets suspended for a short moment, before appearing again and then opening the external editor. While this is the correct behavior for internal editors like vim, it's quite annoying for windowed ones.

## Steps to reproduce

- define a windowed `$EDITOR`, e.g. `code`
- view a diff with tuicr
- in any file, press `e` to open the editor

JammingBen:
fixes https://github.com/agavra/tuicr/issues/580

`e` suspended the terminal around every editor launch, which flickers for editors that open their own window. Windowed editors are now spawned detached and cleaned up as they exit, with an early exit still reported. Terminal editors keep the blocking behaviour, as does any `$EDITOR` carrying `--wait`.

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -122,6 +122,12 @@ Shown below the file tree when local comments or visible remote PR threads exist
 | `e` | Open focused file in `$EDITOR` |
 | `y` | Copy review to clipboard |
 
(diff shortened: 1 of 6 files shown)


agavra (approved):
LGTM, thanks @JammingBen

PR #582, opened 2026-08-10
feat: add a summary command

### LINKED ISSUE #562: Add a command to view pending comments
After I review code, I tend to look at the comments I will submit to decide if I will submit the review as a "comment" or as "request changes."

I would greatly appreciate the ability to do this in `tuicr`!

thedadams:
The summary command shows all the pending comments. Comments can be selected using j/k, and hitting enter on a comment takes you to the comment in the file.

fixes https://github.com/agavra/tuicr/issues/562


src/ui/summary_popup.rs
```diff
@@ -0,0 +1,912 @@
+use std::path::Path;
```
agavra, line 1: thanks for the PR @thedadams I think this is super useful! 

that said, I generally prefer to avoid popup modals in a TUI, I think it's better to just display a full screen (only `?`/`:help` right now is a modal, though even that I'm considering making a full screen) - see #511

thedadams: Thanks for that feedback. Updated the implementation with your feedback.

agavra (approved):

PR #584, opened 2026-08-10
Add settings to skip PR checks and conversation comments

antonio2368:
Adds `show_pr_checks` and `show_pr_comments` config options, both enabled by default.
Set either to `false` to avoid fetching and rendering large GitHub check rollups or PR conversation comments. This improves responsiveness for PRs with large CI output or large comments.

We have a lot of checks and a lot of comments generated by our bots which can be fairly large. This led to quite unresponsive experience when reviewing certain PRs :cry:


src/forge/github/gh.rs
```diff
@@ -171,13 +194,25 @@ where
             default_repository,
             runner,
             local_checkout: None,
+            show_pr_checks: true,
```
agavra: I don't think we ever need pr checks? We can consider just defaulting this to false

antonio2368: I didn't see a need for them in a tool like this.
Should we disable it then?

agavra (approved):

agavra:
Thanks for the contribution @antonio2368 🙏

PR #589, opened 2026-08-11
feat(skills): add a cmux wrapper for launching tuicr

vandamm:
Adds `skills/tuicr/tuicr-wrapper-cmux.sh`, so the skill can open a tuicr review pane in [cmux](https://cmux.com) the same way it already does in tmux, Zellij, and Herdr.

## Behavior

Three intentional differences from the tmux wrapper:

- **No already-running guard.** `cmux list-panes` reports no per-pane command, so the only available check is a global `pgrep -x tuicr`, which would block a legitimate review whenever another cmux workspace has one open.
- **Pass-through tuicr args after `--`** (e.g. `-- -w`, `-- -r <revset>`), which is how an agent scopes the review. Easy to drop if you would rather keep the wrappers identical.
- **Non-blocking.** tuicr is launched with `exec`, so no shell survives it and cmux closes the pane on exit. The wrapper returns as soon as the pane is running and prints the surface ref, which lets an agent poll `tuicr review comments` while the review is still open.

## Skill changes

- new `$CMUX_WORKSPACE_ID` row in the environment table. That variable detects cmux.
- cmux entries in the multiplexer tips and the wrapper path list.
- README and AGENTS.md now say "cmux, tmux, Zellij, or Herdr".

No Rust code touched.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,7 +110,7 @@ src/
 
 Repository-managed agent integrations:
 
-- `skills/tuicr/` - Shared agent skill bundle for coding agents, for example Claude Code, Codex, and similar tools; launches tuicr in a tmux, Zellij, or Herdr split pane
+- `skills/tuicr/` - Shared agent skill bundle for coding agents, for example Claude Code, Codex, and similar tools; launches tuicr in a cmux, tmux, Zellij, or Herdr split pane
 
 ### Key Types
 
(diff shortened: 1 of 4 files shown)


agavra (approved):
I don't use cmux so I'll just trust you on this one 🙏 thanks!

PR #593, opened 2026-08-11
fix: :prs command erroring in azuredevops, fix filter issue for other forges

### LINKED ISSUE #591: :prs command errors for azuredevops
Repro steps:

1. Open `tuicr` in a local azure devops repo that has 2 PRs.
1. Call `git fetch`
1. Tab to view the PR list (note 2 or more prs)
1. Open 1 PR for review
1. command `:prs` to go back to the list of pull requests
1. try to open other PR
1. see error 

Image

Note: need to test if this is only occurring in devops or in other forges.

joshvito:
Here is a summary of the UX that I propose we change.
This fixes #591, and an unreported issue with PR "hopping" and .tuicrignore being "ignored".

---

#### Before: first PR of session opens fine. :prs → pick another PR → red banner:

Failed to open PR #1498: Reviewing an Azure DevOps PR needs a local clone of the
repository (Azure exposes no unified-diff API). Run tuicr from inside a clone of the repo.

PR never opens. Message is misleading — you were inside the clone. Only workaround = quit tuicr, relaunch, one PR per process.

### GitHub user

Nothing was ever fatal — diff comes from GitHub API, so the 2nd PR opened fine. That why UI "works as expected". But 2nd+ PR silently lost the clone, so:

#### Before (2nd+ PR only):
- .tuicrignore not applied — files you deliberately hide (lockfiles, generated code, snapshots) reappear in tree and diff. Looks like ignore file randomly stopped working.
- :e / open-in-editor fails: Cannot open src/foo.rs: no local checkout.
- Context expansion + commit-range diffs lose the local fast path, fall back to gh api — still correct, just network round-trips instead of local reads. Slower, more rate-limit burn.

### Both

Unaffected paths: first PR of a session, and tuicr pr <n> straight from CLI — those already resolved the clone correctly. Bug only bite on PRs opened after entering PR mode, which is exactly the :prs loop.

Fixes #591


agavra (approved):
LGTM! thanks, after fixing the merge conflict I'll merge it

joshvito:
> LGTM! thanks, after fixing the merge conflict I'll merge it

done

agavra:
Thanks @joshvito - sorry I missed your update :)

Issue #76, opened 2026-01-13
shift-G moves to the start of the last file instead of bottom of viewport


agavra:
`shift-G` should move the cursor to the last line, not the first line of the last file

gefersonhess:
I think this got fixed by #240. @agavra

Issue #11, opened 2026-01-08
scroll allows you to go past content


agavra:
we don't limit how far you can scroll right or down (only left/up) so you can scroll so far you don't see anything

airscholar:
@agavra is this still a problem?

PR #595, opened 2026-08-12
fix: keep a zz centering when opening a comment near EOF

davinkevin:
## Problem

`zz` on a line near the end of a file centers the cursor and leaves blank rows below the content. Opening a comment on that line snaps the view back, so the line being reviewed jumps to the bottom of the viewport. `Esc` does not restore the centering.

### Repro

1. Open any diff and move the cursor near the end of a file — far enough that `zz` leaves blank rows below the content.
2. Press `ZZ` to center the line.
3. Press `c` to open a comment on it.

The line drops from the middle of the viewport to near the bottom. It happens in both the unified and side-by-side views.

## Cause

Two places disagree on the maximum scroll offset.

`App::max_scroll_offset` (`src/app/navigation.rs`) explicitly allows blank space below the content, which is what makes `zz` work at EOF:

```rust
/// This permits empty space below content (e.g. when centering the cursor near EOF)
pub fn max_scroll_offset(&self) -> usize {
    self.total_lines().saturating_sub(1)
}
```

`scroll_comment_input_into_view` (`src/ui/diff_view.rs`) clamped with a stricter "no blank space below" rule:

```rust
let max_scroll = total_lines.saturating_sub(viewport_height);
```

That clamp runs on every frame an input box is rendered (`diff_unified.rs`, `diff_side_by_side.rs`). The early `return` only checks whether a box exists, so the clamp fired even when the box was already fully visible and no adjustment was needed.

Worked example — viewport 40 rows, 100-line document, `zz` on line 98:

| step | `scroll_offset` | screen row of line 98 |
| --- | --- | --- |
| after `zz` | 78 (18 blank rows below) | 20 — centered |
| comment opened (4-line box, 104 lines total) | clamped to 104 − 40 = **64** | 34 — near the bottom |

## Fix

Clamp to `total_lines - 1` so the render path and the app agree. The branches that pull the box into view when it is above, below, or taller than the viewport are untouched, so the box still follows the text cursor while typing.

## Tests

- `should_keep_cursor_centered_when_opening_comment_near_eof` — the scenario above; `scroll_offset` stays at 78.
- `should_clamp_to_last_line_when_box_sits_past_the_content` — covers the changed clamp directly.
- `should_not_scroll_past_end_of_content` asserted 90, but that value came from the box-above-viewport branch, not the clamp — the clamp never ran in that case. It now asserts the 95 that branch produces, and the new test above pins the clamp.

`cargo test` (1363 passed), `cargo clippy --all-targets -D warnings`, and `cargo fmt --check` are clean. Also verified interactively with a release build: centering survives at EOF, and the input box still auto-scrolls into view at the top and middle of a file, in both diff layouts.

No issue filed for this — I could not find one open that tracks it (closest is #76, different code path; #11 is likely where the strict clamp came from).

### DIFF
--- a/src/ui/diff_view.rs
+++ b/src/ui/diff_view.rs
@@ -376,8 +376,11 @@ pub(super) fn scroll_comment_input_into_view(
         }
     }
 
-    // Clamp so we never scroll past the last line.
-    let max_scroll = total_lines.saturating_sub(viewport_height);
+    // Clamp so we never scroll past the last line. This must match
+    // `App::max_scroll_offset`, which allows empty space below the content:
+    // clamping to `total_lines - viewport_height` instead would undo a `zz`
+    // centering near EOF every frame the input box is rendered.
+    let max_scroll = total_lines.saturating_sub(1);
     if *scroll_offset > max_scroll {
         *scroll_offset = max_scroll;
     }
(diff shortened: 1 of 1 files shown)


agavra (approved):
nice, this is subtly improved.

Issue #483, opened 2026-07-19
Feature request: mark a comment resolved


khaitranhq:
Please add support for resolving comments in both the TUI and CLI. This is useful for seamless integration with custom skills and automation.

mathstuf:
Note that, in general, there is the "reviewers resolution" and the "contributors resolution". I, personally, dislike Github's handling of comment resolution because it assumes an update to the context is an automatic resolution when I really want verification from the reviewer (for more than "fix a typo" kind of things). I'd like to see that modeled here even if the forges are ignorant of it.

agavra:
this makes sense to me, we'd have to think through how the UX works for it. I'm open to suggestions.

davinkevin:
I'd say anything similar to threads from GitHub/GitLab(/Hunk?) is a good solution IMO. 

It's the (major) missing feature to interact not with human, but just agents, with potential back-and-forth. 🏓

Issue #410, opened 2026-06-07
Support resolving review threads and replying to existing comments


uribrecher:
Problem
tuicr currently supports authoring new line/range/file/review-level comments and submitting them via :submit, but there's no way to interact with existing review threads on a PR — specifically:

Resolving a conversation — no equivalent of GitHub's "Resolve conversation" button.
Replying to an existing thread — c / C create new top-level comments rather than threading a reply onto an existing comment.
For PR authors responding to review feedback (or reviewers following up on a thread), this means dropping back to the GitHub web UI.

Proposal
Render existing review threads inline next to the relevant hunk, with thread state (open/resolved).
Add a keybinding to focus a thread (e.g. navigate threads with [t / ]t).
On a focused thread:
A reply action (e.g. a for "reply") that opens the comment editor and posts via addPullRequestReviewThreadReply.
A resolve/unresolve toggle (e.g. R when a thread is focused, or :resolve) that calls the resolveReviewThread / unresolveReviewThread GraphQL mutations.

API references
resolveReviewThread / unresolveReviewThread GraphQL mutations
POST /repos/{owner}/{repo}/pulls/{pr}/comments/{comment_id}/replies REST endpoint

agavra:
Agreed! #337 is already on the todo list. Resolving conversations is a good idea.

sylvain-chaugny-assessio:
Can't wait for this! :D

jonathanmorris678:
+1, would love this feature!

pitoniak32:
I would love to be able to reply to a thread of comments, or possibly filter comments from specific users (copilot can get noisy)

Lackmello2345-creator:
Sent from my iPhone

hacooke:
I wouldn't mind taking a run at these features if nobody has started this yet. I will try to get a PR up in the near future if that's okay

PR #599, opened 2026-08-13
feat(gitlab): resolve, reply to, edit and delete MR discussion threads

samuelarbibe:
Addresses #410 and #483 for GitLab. Opening it as a concrete proposal rather than a finished answer for every forge — the trait is the general part, the GitLab backend is the one implementation, and the UX is deliberately small. Happy to reshape any of it.

## Problem

Remote threads are read-only. `i` and `dd` on one answer `read only in tuicr`, and `c` refuses because the cursor is not on a diff line. Responding to review feedback means leaving for the web UI — which, for the agent-driven back-and-forth #483 describes, is the whole loop.

## What this adds

`ForgeBackend` gains four thread mutations behind a `supports_thread_mutations()` capability check, each defaulting to an unsupported-operation error. GitHub, Bitbucket and Azure are untouched and keep today's behaviour; the UI asks the capability before offering an operation, rather than letting you compose a reply and rejecting it on submit.

GitLab implements them over the discussions REST endpoints. Bodies go out as `--raw-field` so text containing `=`, or text that looks like a filename, survives `glab` intact.

In the diff view, with the cursor on a thread:

| Input | Result |
|-------|--------|
| `:resolve` / `:unresolve` | Toggle the thread's resolved state |
| `:reply`, or `c` on a thread | Open the editor; Ctrl-S posts the text as a reply |
| `i` | Edit the note under the cursor |
| `dd` | Delete the note under the cursor |

`c`/`i`/`dd` keep their existing meanings everywhere else — the thread path is only reached where those keys previously printed a refusal.

## On the two design questions in the issues

**Ownership.** Editing and deleting are gated on the login the forge reports for the viewer, not the local `username` config, so someone else's note is refused inside tuicr rather than failing at the API. No reported identity means both are refused.

**Resolution semantics.** @mathstuf's point in #483 — that a reviewer's resolution and a contributor's are different things — is real, but modelling it beyond what the forge stores would need state tuicr does not have. This maps `:resolve` straight onto the forge's own resolved flag and nothing more, which leaves room for a richer model later.

Each write is a single request followed by a refetch, so the view shows what the forge holds rather than a locally patched guess.

## Tests

Ten new tests: four in the GitLab backend pinning the exact `glab` argv for each endpoint, and six driving the app through the cursor — resolve on and off a thread, a reply typed into the editor, editing your own note, refusing someone else's, refusing when the backend cannot mutate threads, deleting your own, and refusing when the forge reports no viewer identity.

`cargo test` (1374 passed), `cargo fmt --check` and `cargo clippy --all-targets` are clean. `docs/GITLAB.md` gains a "Discussion threads" section.

Verified against a self-hosted GitLab instance.

### DIFF
--- a/docs/GITLAB.md
+++ b/docs/GITLAB.md
@@ -58,6 +58,26 @@ not add you as a reviewer on your behalf.
 `:submit draft` remains GitHub-only. Run it against a GitLab MR and tuicr returns
 an unsupported-operation error rather than submitting.
 
+## Discussion threads
+
+Existing MR discussions are writable from the diff view. Put the cursor on a
+thread, then:
+
+| Command | Result |
+|---------|--------|
+| `:resolve` / `:unresolve` | Toggles the thread's resolved state. |
+| `:reply`, or `c` on a thread | Opens the editor; Ctrl-S posts the text as a reply. |
+| `i` | Edits the note under the cursor. |
+| `dd` | Deletes the note under the cursor. |
+
+Editing and deleting are limited to your own notes. Ownership is checked against
+the login GitLab reports for your token, so someone else's note is refused in
+tuicr rather than failing at the API. When the forge reports no identity at all,
+both are refused.
+
+Each write is a single request and the thread list is refetched afterwards, so
+the view shows what GitLab now holds rather than a locally patched guess.
+
 ## Self-hosted GitLab
 
 Authenticate against your instance once:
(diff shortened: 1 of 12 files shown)


agavra:
Thanks @samuelarbibe - I'm going to punt this PR for now, I'd like to design this flow myself since it's pretty core to the tuicr experience. If I don't get to it in the next few weeks I'll battle test your patch.

PR #600, opened 2026-08-13
perf(comments): cull comment boxes outside the viewport

### LINKED ISSUE #552: TUI lag scales with total comment volume in a session, not with what is on screen
> Issue created with the assistance of Claude Opus 5.

## Summary
Once a review session accumulates enough comment text, the whole TUI becomes sluggish. This is not limited to typing a reply - scrolling, navigation, and general redraw are all affected, and the cost is paid for comments that are scrolled far off screen.

Comment bodies are re-highlighted through syntect's Markdown grammar on every render frame, with no memoization, and the renderer's viewport culling does not cover the comment branch. So per-frame cost grows with the total amount of comment text in the session rather than with what is visible.

## Symptoms

Scrolling and navigation become visibly sluggish across the whole diff, not only near a large comment. Typing in the comment box is the worst case, because comment-input mode disables viewport culling entirely. Deleting large comments improves responsiveness incrementally, and the improvement tracks how much total comment text was removed.

## Measurements

Bodies were posted with `tuicr review add` and viewed in the TUI. "Spans" counts inline markup, where one backtick pair or one `**` pair is one span. Comments were added one at a time and never removed, so the session total grew monotonically.

| Comment size | Spans in that comment | Session total (chars) | Observed |
|-------------:|----------------------:|----------------------:|----------|
| 1685 | 0 | 9,621 | smooth |
| 1691 | 8 | 11,312 | smooth |
| 1641 | 8 | 12,953 | smooth |
| 1629 | 16 | 14,582 | smooth |
| 3372 | 0 | 17,954 | noticeably slower |
| 3444 | 64 | 21,398 | heavy lag |

I originally read this as a per-comment effect, but that interpretation is confounded: each new comment also increased the session total, so per-comment size and session total moved together. Removing comments afterwards is what separated them. Deleting the two largest comments improved responsiveness step by step, and with only a single ~1700 char comment of my own left the interface was still sluggish, because roughly a dozen earlier comments totalling around 23,000 characters were still loaded.

So the dominant variable is total comment text in the session. Per-comment size and markup density matter mainly through their contribution to that total.

Perception is coarse and I did not instrument frame times, so treat the thresholds as indicative rather than exact.

## Repro

```bash
cat > /tmp/genbody.py <<'PYEOF'
import sys
targetChars, codeSpans, boldSpans = int(sys.argv[1]), int(sys.argv[2]), int(sys.argv[3])
words = []
for i in range(codeSpans):
    words.append("`identifierName%d`" % i)
for i in range(boldSpans):
    words.append("**emphasis%d**" % i)
out, i = [], 0
while len(" ".join(out)) < targetChars:
    if i < len(words):
        out.append(words[i]); i += 1
    else:
        out.append("word")
    if len(out) % 12 == 0:
        out.append("\n")
sys.stdout.write(" ".join(out))
PYEOF

SESSION='<your-session-slug>'
REPO="$(pwd)"
FILE='<a file in the diff>'
post() {
  tuicr review add --repo "$REPO" --session "$SESSION" --target-file "$FILE" \
    --line 1 --side new --type note --username repro "$(python3 /tmp/genbody.py "$@")"
}

# Add ~3.4k chars of comment text at a time and scroll after each.
for i in $(seq 1 8); do post 3400 37 27; done
```

Scroll through the diff after each addition. Degradation is progressive. Then scroll to a part of the file with no comments at all and confirm it is still slow, which is the part that shows culling is not applying to comments.

## Environment

- `tuicr` 0.19.1, installed via `cargo install`. Source references below are from v0.20.0; those code paths look unchanged, but I have not measured on 0.20.0.
- tmux 3.4, `TERM=tmux-256color`, Linux.

## Suspected cause

`src/ui/comment_panel.rs`, `markdown_body_lines` clones the body and highlights it on every call, and `format_comment_lines` calls it unconditionally:

```rust
let owned: Vec<String> = lines.iter().map(|s| s.to_string()).collect();
let highlighted = theme.syntax_highlighter().highlight_markdown_lines(&owned);
```

`src/syntax/mod.rs`, `highlight_lines_with` builds a fresh `HighlightLines` state machine per call and allocates an owned `String` per style run:

```rust
let mut highlighter = HighlightLines::new(syntax, &self.theme);
...
.map(|(style, text)| (Self::syntect_to_ratatui_style(style), text.to_string()))
```

`SyntaxHighlighter` holds only `syntax_set`, `theme`, `add_bg` and `del_bg`. There is no memo map and no interior mutability. By contrast `DiffLine` already carries a `highlighted_spans` cache populated once at load, so diff lines do not pay this cost per frame. Comments have no equivalent.

### Why off-screen comments still cost

In `src/ui/diff_unified.rs::render_unified_diff`, the `visible_start` / `visible_end` guard wraps only the diff-line span building. The comment branch below it runs unconditionally, so every comment in the review is re-highlighted every frame regardless of scroll position. This is the part that makes the cost scale with session totals rather than with the viewport.

### Why typing is the worst case

`src/ui/diff_view.rs::diff_visible_range` returns an unbounded range in comment-input mode:

```rust
if app.input_mode == crate::app::InputMode::Comment {
    (0, usize::MAX)
} else {
    ...
}
```

So while typing, every frame rebuilds the entire document on top of the already-uncached comment highlighting.

I have not profiled this, so treat it as a hypothesis grounded in reading the source rather than a measured diagnosis. Line numbers were counted from the raw blobs at v0.20.0 and may be off slightly; function names are exact.

## Possible fix

Cache highlighted output for saved comments, keyed on content hash plus theme. Comment bodies are immutable once saved, so the per-frame work is entirely redundant. Caching `HighlightedLines` on the comment, mirroring the existing `DiffLine.highlighted_spans` pattern, would fit the codebase, and `src/hash.rs` with `DiffFile::compute_content_hash` is established precedent. Only the live `format_comment_input_lines` buffer genuinely changes per keystroke.

Extending the viewport guard to cover the comment branch would help independently, and would also reduce the cost of the input-mode case.

## Related but distinct

- #278 (closed) and #58 (open) concern diff load performance. 
- #288 and #394 (open) move diff highlighting to a background worker.
- #26 (closed) established performance versus highlighting as a known concern. 

None of these touch comment-body markdown highlighting.

joshvito:
## Summary

Cull comment boxes outside the viewport in both diff renderers and in the PR issue-comments section, substituting the exact number of blank rows. Per-frame cost tracked total comment volume in the session rather than what was on screen; it is now roughly flat in comment count.

Fixes #552.

## Root cause

The renderer's visible-range guard only wrapped diff-line span building. The comment branch below it ran unconditionally, so a comment box 3,000 rows off screen was formatted, wrapped and span-built in full before being discarded.

## Measurements

`cargo test --release render_perf -- --ignored --nocapture`, same machine, `main` at 5ec9a83.

**Comment count** (20 files x 200 lines):

| comments | main | this PR |
| ---: | ---: | ---: |
| 0 | 0.65 ms | 0.49 ms |
| 20 | 1.80 ms | 0.62 ms |
| 60 | 4.01 ms | 0.82 ms |
| 200 | 12.74 ms | 1.04 ms |

The 12x at 200 comments is not the point — the shape is. Cost was linear in total comment volume; the residual growth on this branch is row bookkeeping and comment bars, not formatting.

**Diff size** is unchanged, as expected — nothing here touches diff-line rendering:

| diff lines | main | this PR |
| ---: | ---: | ---: |
| 200 | 0.35 ms | 0.37 ms |
| 40 000 | 2.48 ms | 2.57 ms |

## Design notes

- **Row counts for culled boxes come from `App::comment_display_lines`**, the mirror the annotation builder already uses, so this is a second consumer of an existing invariant rather than a new one. `comment_display_lines_matches_rendered_box_height` pins that mirror against `format_comment_lines` across bodies, multibyte text, and widths narrower than the box chrome.
- **Comment bars are still recorded for culled boxes**, since a bar is painted above its box and can be on screen while the box is not.
- **Not culled, deliberately:** comment-input mode (the scroll offset can move after the document is built, and #628's parser keeps typing cheap anyway) and remote review threads (plain spans, no markdown pass).

## Verify

`should_cull_comment_boxes_outside_the_viewport` in both renderers scrolls to the row `line_annotations` assigned the comment and expects the box there — if the placeholder count were off by even one, that index would point somewhere else.

No dependency or build changes: `Cargo.toml` and `Cargo.lock` are identical to `main`.

### DIFF
--- a/src/ui/comment_panel.rs
+++ b/src/ui/comment_panel.rs
@@ -584,6 +584,55 @@ mod tests {
         Theme::default()
     }
 
+    /// The renderer replaces off-screen comment boxes with exactly
+    /// `App::comment_display_lines` blank rows, and the annotation builder sizes
+    /// every comment the same way. If that count ever drifted from what
+    /// `format_comment_lines` actually emits, the document would desync — the
+    /// cursor would land on the wrong row and culled boxes would leave the wrong
+    /// number of gaps. Pin the two together.
+    #[test]
+    fn comment_display_lines_matches_rendered_box_height() {
+        let theme = test_theme();
+        let bodies = [
+            "",
+            "single line",
+            "first\nsecond\nthird",
+            "trailing newline\n",
+            "\n\nleading blanks",
+            &"x".repeat(300),
+            &"日本語のテキストです ".repeat(20),
+            "`code` **bold** and a very long tail that will need to wrap at least once or twice",
+        ];
+        // Viewport widths, including degenerate ones narrower than the box chrome.
+        for viewport_width in [9usize, 12, 40, 80, 120] {
+            for body in bodies {
+                let comment = crate::model::Comment::new(
+                    body.to_string(),
+                    crate::model::CommentType::from_id("note"),
+                    None,
+                );
+                let rendered = format_comment_lines(
+                    &theme,
+                    CommentTypePresentation {
+                        label: "NOTE".to_string(),
+                        color: Color::Blue,
+                    },
+                    &comment.content,
+                    None,
+                    // What every call site passes: the viewport minus the
+                    // cursor-indicator column.
+                    viewport_width.saturating_sub(1),
+                    None,
+                );
+                assert_eq!(
+                    App::comment_display_lines(&comment, viewport_width),
+                    rendered.len(),
+                    "width={viewport_width} body={body:?}"
+                );
+            }
+        }
+    }
+
     // -- wrap_segments tests --
 
     #[test]
--- a/src/ui/diff_side_by_side.rs
+++ b/src/ui/diff_side_by_side.rs
@@ -17,7 +17,7 @@ use crate::ui::diff_view::{
     apply_horizontal_scroll, comment_type_presentation, cursor_indicator, cursor_indicator_spaced,
     diff_stat_title, hunk_header_text_and_style, paint_cursor_line_highlight,
     paint_visual_selection_overlay, populate_row_to_annotation, render_expander_line,
-    render_hidden_lines, scroll_comment_input_into_view,
+    render_hidden_lines, scroll_comment_input_into_view, skip_comment_box,
 };
 use crate::ui::styles;
 use crate::ui::text_utils::{
@@ -204,6 +204,11 @@ impl SideBySideContext<'_> {
         line_idx >= self.visible_start && line_idx < self.visible_end
     }
 
+    /// Same question for a multi-row comment box: does any of it land on screen?
+    fn box_visible(&self, top: usize, rows: usize) -> bool {
+        crate::ui::diff_view::comment_box_visible(top, rows, (self.visible_start, self.visible_end))
+    }
+
     fn search_for(&self, line_idx: usize) -> Option<(&str, Style)> {
         let needle = self.app.search_paint_at(line_idx)?;
         Some((needle, self.search_style))
@@ -369,6 +374,11 @@ pub(super) fn render_side_by_side_diff(frame: &mut Frame, app: &mut App, area: R
                 line_idx += 1;
             }
         } else {
+            let rows = App::comment_display_lines(comment, ctx.panel_width);
+            if !ctx.box_visible(line_idx, rows) {
+                skip_comment_box(&mut lines, &mut line_idx, rows);
+                continue;
+            }
             let comment_lines = comment_panel::format_comment_lines(
                 &app.theme,
                 comment_type_presentation(app, &comment.comment_type),
@@ -456,6 +466,7 @@ pub(super) fn render_side_by_side_diff(frame: &mut Frame, app: &mut App, area: R
         &mut line_idx,
         ctx.current_line_idx,
         ctx.panel_width.saturating_sub(1),
+        (ctx.visible_start, ctx.visible_end),
     );
 
     for (file_idx, file) in app.diff_files.iter().enumerate() {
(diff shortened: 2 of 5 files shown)


agavra:
@joshvito let me know when this one is ready to go, if you can update the description with what the final approach was that'd be great

joshvito:
After re-running the harness, the difference in render time is much less. 
I can understand if we just decide to defer this work and abandon the branch. 

I think the replacement of the double regex package is worth doing even if we don't merge the rest of this code change (happy to put the package change in a separate branch, just say the word).

**EDIT**: re-ran the harness with only the package change, and it was a regression. going to revert that change from this branch.

Icohedron:
Thanks for this PR!

I built tuicr from this PR's branch and it works wonders on my machine. 
tuicr went from being nigh unusable for me, to now being my go-to review tool.
If there are any issues, I have yet to encounter them. I will comment if I do encounter any.

I look forward to the day these features get merged :+1:

joshvito:
@agavra I am still using this branch as a daily driver, and see no related issues to these changes. However, we should make a decision if the changes are worth the performance improvements.

agavra:
Thanks @joshvito - can we split this into two PRs? The first for the fix that only renders visible comments and the second for the caching? It seems like the first is sufficient for 99% of the use cases and is a no-brainer to me, caching the result of rendering seems acceptable but I'd like to think about it a little bit more. I'm always anxious to introduce a cache, but I admit that caching based on content for markdown seems pretty foolproof...

joshvito:
Sure, I'll take a look in the morning, and split the changes out into more
manageable PR.

agavra:
https://github.com/agavra/tuicr/pull/628 - I played around with this, it removes the need for the memoization by changing the parser from regex to something much better. so let's jsut repurpose this PR for the culling optimization. culling + #628 should make comment rendering go brrr

thanks for pushing on this @joshvito!

agavra (approved):
thanks! these two comment improvements are going to make a big difference :)

Issue #602, opened 2026-08-13
Parentheses in file names throw error when using jj backend


BennettQuaritsch:
When a file has parentheses in the filename, an error is thrown on `tuicr` invocation.

```bash
❯ tuicr 
Error: VCS command failed: jj file show -r @- -T "\n@@TUICR_BATCH_BOUNDARY_e97f2d44_8b1a@@\n" ++ path ++ "\n" frontend/src/lib/components/RecipeThumb.svelte frontend/src/lib/features/meal-plans/components/DayCard.svelte frontend/src/lib/features/meal-plans/components/Slot.svelte frontend/src/lib/features/meal-plans/components/SlotActionSheet.svelte frontend/src/lib/features/meal-plans/components/TodayCard.svelte frontend/src/lib/features/meal-plans/components/TodaySection.svelte frontend/src/lib/features/meal-plans/components/WeekGrid.svelte frontend/src/lib/features/meal-plans/components/WeekSwitcher.svelte frontend/src/routes/(app)/+page.svelte failed: Error: Failed to parse fileset: Syntax error
Caused by:  --> 1:21
  |
1 | frontend/src/routes/(app)/+page.svelte
  |                     ^---
  |
  = expected <EOI>, `|`, `&`, or `~`
Hint: See https://docs.jj-vcs.dev/latest/filesets/ or use `jj help -k filesets` for filesets syntax and how to match file paths.
```

agavra:
oof that's a bad one! thanks for the report, I haven't used tuicr on much nextjs stuff

PR #603, opened 2026-08-13
Fix parenthesis in filenames throwing an error when using jj backend

BennettQuaritsch:
When using jj backend, using parenthesis like src/routes/(app)/+page.svelte would break tuicr. Filenames should be in parentheses (and escaped for safety).

See issue #602. The problem in this issue is fixed with this. Additionally escaping of " and \ characters is added.


BennettQuaritsch:
Fixed in #619

PR #607, opened 2026-08-14
Add configurable file line stats

antonio2368:
Show GitHub-style `+added -removed` counts in the file list. Add `show_file_line_stats` to disable them when needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,7 @@ Path: `~/.config/tuicr/config.toml` on Linux/macOS, `%APPDATA%\tuicr\config.toml
 ```toml
 theme = "catppuccin-mocha"
 diff_view = "side-by-side"   # or "unified"
+show_file_line_stats = true  # show per-file +added -removed counts
 ignore_whitespace = false    # ignore all whitespace in local VCS diffs
 appearance = "system"        # or "dark" / "light"
 mouse = true
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -27,6 +27,7 @@ theme_light = "gruvbox-light"
 diff_view = "side-by-side"
 ignore_whitespace = false
 show_file_list = true
+show_file_line_stats = true
 show_pr_checks = false
 show_pr_comments = true
 mouse = true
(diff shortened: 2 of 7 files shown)


ac8318740:
Had a similar thought and really like this idea. I'd built a version myself before noticing yours, so I'm not going to submit a duplicate – a few notes on yours instead, plus the couple of things I did differently.

**Pristine mode shows `+0 -0` on every row.** `--all-files` and `--file` review unchanged code, so every file stats as `+0 -0`. The row builder already suppresses the M/A/D badge there for exactly this reason – the existing comment reads "Pristine mode reviews unchanged code; the M/A/D badge would lie" – and the counts have the same problem.

Correcting myself here: I first suggested guarding on `!app.is_pristine_mode` alone, and that only fixes half of it. `--all-files` sets that flag, but `--file` leaves it false and goes through `FileBackend::new` with `vcs_type == VcsType::File`, so it keeps rendering the counts. `diff_watch_tick` already pairs the two conditions for the same underlying reason. A predicate covering both reads better than repeating the pair:

```rust
// on App
pub fn is_whole_file_view(&self) -> bool {
    self.is_pristine_mode || self.vcs_info.vcs_type == VcsType::File
}
```

then `app.show_file_line_stats && !app.is_whole_file_view()` at both call sites in `file_list.rs`. I've been running that locally against your branch and it behaves – counts gone in both whole-file modes, unchanged for normal diffs.

**Pure renames also render `+0 -0`.** I dropped the suffix when both counts are zero. GitHub does show 0/0 on a rename, so this is a preference rather than a bug – raising it in case you'd rather stay consistent with the forge.

**The `is_commit_message` guard is a good catch**, I'd missed it. Worth noting it only earns its keep in `file_stat_width`: `file_stat_spans` never sees a commit-message row, since that branch builds its own `Line` and returns above the call site.

**Default.** You default on, I defaulted off. On is friendlier and I don't feel strongly – only flagging it in case a default-off opt-in turns out to be an easier sell for a change every user sees.

Happy to send any of this as a patch if it saves you time.

antonio2368:
@ac8318740 thanks for all the suggestions, they sound good to me!
Can you send the patch so I don't try to do the same work, it's probably faster for you to just apply it directly.
I'm not stuck on doing things in the same way as some other services, so I'm okay to handle renames in a way you feel is more readable.

PR #608, opened 2026-08-14
fix(diff): decode C-quoted non-ASCII paths in diff headers

mosugi:
With `core.quotepath` at its default, Git octal-escapes every non-ASCII byte in the paths it writes into diff headers and wraps them in double quotes, so `日本語.txt` arrives as
`"\346\227\245\346\234\254\350\252\236.txt"`. The parser stripped the `a/`/`b/` prefixes without unquoting first, and because the leading quote is in the way the prefix was not even stripped: the escapes, quote and all, became the file's path. They showed up in the file tree and diff headings, and were what comments got exported against. Anyone reviewing a repo with Japanese (or any non-ASCII) filenames saw escape sequences instead of names.

Decode C-style quoting on every path the parser reads: the `diff --git` header, `---`/`+++`, `rename`/`copy from`/`to`, and the binary-file line. Git quotes each side of the `diff --git` header independently, so a rename between a non-ASCII and an ASCII name mixes a quoted and a bare path; split that header on the quoting rather than on `" b/"` alone. Paths whose bytes are not valid UTF-8 keep their escapes, which is lossless.

Git also terminates a `---`/`+++` path with a tab when the name contains a space, which left the closing quote mid-line and defeated unquoting. That suffix is now stripped for Git as well as Mercurial; an unescaped tab is never part of a name, since a real one forces quoting.

This affects the Git CLI backend (sparse checkouts and `backend = "cli"`), jj, hg, and patches fetched from a forge — the libgit2 backend gets its paths from the library, not from diff text.

### DIFF
--- a/src/vcs/diff_parser.rs
+++ b/src/vcs/diff_parser.rs
@@ -57,7 +57,7 @@ where
     // `next_line` instead of `lines.next()` to propagate I/O errors.
     while let Some(line) = next_line(&mut lines)? {
         if line.starts_with(header_prefix) {
-            let (mut old_path, mut new_path, status) = parse_file_header(&mut lines, format)?;
+            let (mut old_path, mut new_path, status) = parse_file_header(&mut lines)?;
 
             // For git-style diffs (jj, git patches), if parse_file_header didn't find
             // ---/+++ or rename/copy lines (e.g. empty new files, mode-only changes),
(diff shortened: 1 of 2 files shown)


agavra:
Superceded by #602 - thanks for the patch!

PR #611, opened 2026-08-15
fix(forge): use the comment type label in the submit prefix

### LINKED ISSUE #659: Configured comment type label is dropped when posting a comment — TUI shows [💬 NOTE], GitHub gets [NOTE]
A comment type with a `label` renders that label in the TUI badge and the markdown export, but the comment posted to GitHub uses the raw id instead.

**Repro**

```toml
[[comment_types]]
id = "note"
label = "💬 note"
```

Write a `note` comment on a GitHub PR and submit it. The TUI badge and `:export` both show `[💬 NOTE]`; the comment that lands on the PR says `[NOTE]`.

**Expected:** the tag on the posted comment matches what the TUI and export show.

This isn't GitHub-specific — it's the shared submit path, so GitLab, Azure DevOps and Bitbucket post the id too.

**The part that actually bites**

When a comment can't be anchored to a diff line, the resolver modal asks you to choose Move-to-summary vs Omit, and it shows the comment with its tag. That preview is built the same way, so it shows `[NOTE]` while the body that gets posted says `[💬 NOTE]` — you're deciding against a tag that isn't what lands. Untyped comments show a literal `[NONE]` there too, where every other resolver shows no tag at all.

`docs/CONFIG.md` documents the tag as one thing prepended "on submit **or** export", and `export_comment_type_label` already exists, so posting the id looks like an oversight rather than intent.

PR #611 fixes this.

barspielberg:
A comment type with a configured `label` shows that label in the TUI badge and the markdown export, but submits the raw id. With `{ id = "note", label = "💬 note" }`, the TUI and export both show `[💬 NOTE]` while the posted comment says `[NOTE]`.

`CommentType` carries only an id — the `label` lives on the config-resolved `CommentTypeDefinition`, so `CommentType::as_str()` can only ever yield the uppercased id. The UI and export paths already resolve through the definitions (`App::comment_type_label`, `export_comment_type_label`); the submit path took a bare `&ForgeConfig`, which holds just `comment_type_prefix`, so it had no way to reach them.

## Change

`build_inline_body` / `map_comment` / `map_range` / `build_review_body` now take a `SubmitContext` (forge config + resolved types) and resolve the label the same way the other two paths do.

`describe_row` in the unmappable-comment resolver modal was a fourth site building the tag from the same enum. It previewed `[NOTE]` while the submitted body said `[💬 NOTE]`, so you pick Move-to-summary vs Omit against a tag that isn't what gets posted. It also printed a literal `[NONE]` for untyped comments, where the other resolvers render no tag at all. Both now match.

`docs/CONFIG.md` said `label` applies to the UI and export; it now says submit too. `AGENTS.md` gets an entry listing the four resolvers that have to agree, since `describe_row` is the easy one to miss.

## Notes

This changes submitted output for anyone already configuring an emoji `label` — the tag becomes what they see in the TUI. The Defaults section of `docs/CONFIG.md` describes the tag as one thing prepended "on submit **or** export", and `export_comment_type_label` already existed, so this reads as the intended behaviour rather than a policy change. Nothing in the repo records an id-only rule.

319 added lines, ~230 of them tests; the production change is under 90.

Four tests added, covering the emoji label on submit, the review body, and the two resolver-modal fixes. CI is green. Two `vcs::git` tests fail on my machine on `main` as well — local git rejects the `relativeworktrees` extension — so they're unrelated to this change.

Closes #659.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -315,6 +315,8 @@ These are non-obvious things the implementation chain hit. Worth preserving for 
 
 16. **GNU Linux release binaries must stay dynamically linked.** Static glibc binaries can crash when hostname lookup loads a host NSS module (for example Fedora's `libnss_myhostname`). The musl artifacts are the supported static Linux builds. Direct updates must preserve the running binary's GNU/musl target environment when selecting an asset.
 
+17. **The `[TYPE]` prefix uses the config-resolved `label`, not the `CommentType` id.** `CommentType` carries no label — it lives on `CommentTypeDefinition` (`App::comment_types`), so `as_str()` only ever yields the uppercased id. Four resolvers render this tag and must agree: `App::comment_type_label` (TUI), `export_comment_type_label` (`src/output/markdown.rs`), `SubmitContext::type_label` (`src/forge/submit.rs`), and `describe_row` (`src/ui/submit_modals.rs`). The last is easy to miss and previews the tag the user is deciding about, so disagreement means choosing Move-to-summary vs Omit against a tag that never gets posted. This is why the submit path takes a `SubmitContext` rather than a bare `&ForgeConfig`.
+
 ### Keeping Docs Updated
 
 When adding user-facing features, update the relevant documentation:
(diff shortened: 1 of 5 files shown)


agavra (approved):
Thank you @barspielberg for the contribution! This looks good to me.

PR #612, opened 2026-08-17
fix(jj): remove Git staging concepts from local selector

0xferrous:
Do not classify the Jujutsu working-copy diff as unstaged changes. The working copy is already represented by @, and jj has no staging index.

Use jj's standard "(no description set)" text for empty commit descriptions.

after/before
image


agavra (approved):
thanks! I don't use `jj` so I appreciate UX improvements there :)

PR #614, opened 2026-08-18
docs: mark Hunk as a standalone binary

benvinegar:
Hunk is available as a standalone compiled binary through Homebrew, mise, Nix, and release artifacts, so it should receive a checkmark in the “Single static binary” row.

Its npm installation does require Node.js, but that is an installation-path detail rather than a runtime requirement for Hunk’s binary distributions.

Please and thanks!


agavra (approved):
Thanks for correcting the inaccuracy! Definitely feel free to update any other part of the matrix if I got stuff wrong.

benvinegar:
Thanks! tuicr looks great

PR #615, opened 2026-08-18
feat(forge): add Forgejo and Codeberg pull request discovery

kennethdsheridan:
## Summary

- add an explicit Forgejo forge type and stable `fj` session slug prefix for Forgejo and Codeberg PR sessions
- recognize Codeberg by hostname and self-hosted Forgejo only when the host exists in `fj`'s key store
- load PR metadata and cumulative diffs through Forgejo REST using `fj` credentials when available

## Safety

- Forgejo and Codeberg review submission, commit selection, commit-range diffs, remote comments, and remote context expansion still fail closed with actionable messages until those API paths are implemented
- GitHub remains the final catch-all parser so self-hosted Forgejo is not guessed from arbitrary hostnames
- use `:clip` to export Forgejo or Codeberg reviews in the interim

## Relationship to Gitea

This PR no longer uses `tea` or labels Gitea-compatible API access as Forgejo support. #652 is the fuller Gitea lane. This branch keeps Forgejo and Codeberg separate with their own `fj` identity.

Release-please impact: minor

## Tested

- `cargo fmt --all --check`
- `cargo test`
- `cargo clippy --all-targets -- -D warnings`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,9 +49,8 @@ src/
 │   └── jj/              # Jujutsu backend (always compiled)
 │       └── mod.rs       # JjBackend: uses jj CLI, parses with diff_parser::GitStyle
 │
-├── forge/               # Remote forge integration (GitHub PR, GitLab MR, Bitbucket PR review)
-│   ├── mod.rs           # Detect and parse GitHub/GitLab/Bitbucket remotes
-│                        # (parse_any_remote_url: Bitbucket, then GitLab, then GitHub)
+├── forge/               # Remote forge integration (GitHub, Forgejo/Codeberg, GitLab, Bitbucket)
+│   ├── mod.rs           # Detect and parse forge remotes; Forgejo discovery uses Tea logins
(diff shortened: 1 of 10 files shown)


onlyhavecans:
Why use the Gitea tea client over the Forgejo cli client [fj](https://codeberg.org/forgejo-contrib/forgejo-cli)?
Tea's ability to manage Forgejo is breaking with every api divergence.

(I was looking to implement this functionality and saw this pr)

kennethdsheridan:
That is a fair criticism. I used tea because it provided a quick authenticated path for Forgejo compatible API discovery, open, and diff, but labeling that as Forgejo support is not the right long term boundary if Forgejo and Gitea continue diverging.

Given #652 now has a fuller Gitea implementation, I am going to keep this draft from moving forward as is. The next version should either use fj for Forgejo and Codeberg or otherwise make Forgejo specific behavior explicit, with a separate fj session identity instead of sharing the Gitea and tea shaped path.

PR #617, opened 2026-08-18
perf(tui): coalesce queued input events into one repaint

### LINKED ISSUE #616: Event loop repaints once per input event, so a wheel gesture renders frames nobody sees
> Rewritten. This issue originally described comment bodies being markdown-highlighted every frame, which duplicates #552 (already open, with #600 open against it). Measurements from that investigation are now on #552, and this issue covers only the separate event-loop defect found alongside it.

The event loop draws one frame per input event, with no coalescing. A trackpad gesture or a held key delivers events faster than a terminal can present frames, so most of those frames are rendered and written to the terminal but never seen.

## What happens

`src/main.rs` polls, reads exactly one event, handles it, and falls through to a repaint:

```rust
if event::poll(Duration::from_millis(100))? {
    needs_redraw = true;
    let event = event::read()?;
    // ... handle one event ...
}
```

Each repaint is bracketed in a synchronized-output pair and flushed. At 180x50 a full repaint is ~20 KB of escape sequences and a diff scroll is ~14 KB, so one wheel gesture can push hundreds of KB for frames that are never perceived.

## Symptoms

The user-visible effect is an ordering that tracks events-per-gesture rather than cost-per-frame:

- mouse wheel / trackpad is the worst, since one gesture is many events (and `WHEEL_LINES` is 3, so each event also scrolls 3 lines)
- held `j`/`k` is better, one event per key-repeat
- a single `m`/`M` jump feels instant, one event

This is only noticeable when a frame is expensive for some other reason. It amplified #552 for me, which is how I found it.

## Measurement

Instrumenting the loop while scrolling a PR, logging per frame whether input was already queued at flush time (`event::poll(Duration::ZERO)` right after the flush): **input was already waiting on 91 of 121 frames**.

Driving tuicr in a real pty and sending 20 keys with no pacing, counting frames by splitting output on the DEC 2026 sync-update marker:

| | frames | bytes |
| --- | --- | --- |
| one frame per event | 22 | 30.8 KB |
| coalesced | 1 | 20.7 KB |

## Fix

Consume every already-queued event before repainting. The first poll still blocks for 100 ms so an idle TUI costs nothing, and a burst is capped so continuous input keeps showing progress rather than sitting on a stale frame.

PR #617.

## Out of scope

- Per-frame byte volume. ~1.6 bytes/cell for a full repaint is near the floor, since per-row diff backgrounds mean a scroll legitimately changes every cell.
- Comment-body highlighting cost: #552 / #600.
- The O(total diff) line rebuild in the renderer: real but small, ~1 ms per 40 000 diff lines.

Environment: tuicr 0.22.0, macOS 15.6, Ghostty, release build.

KallivdH:
Fixes #616.

> Rewritten. This PR originally also cached comment-body markdown highlighting, which duplicates #552 and the work in #600. That commit is dropped: #600 handles it more thoroughly (text-keyed cache with LRU and a byte budget, the syntect `default-fancy` backend instead of onig, and viewport culling for comments). What remains is the separate event-loop defect and the measurement harness that made both visible.

## `perf(tui): coalesce queued input events into one repaint`

The loop drew one frame per input event. A trackpad gesture or held key delivers events faster than a terminal can present frames, so the extra frames were rendered and written but never seen. At 180x50 a full repaint is ~20 KB of escape sequences and a diff scroll ~14 KB, so a single gesture could push hundreds of KB of output nobody perceives.

Instrumenting the loop while scrolling a large PR showed input already queued at flush time on **91 of 121 frames**.

Now every already-queued event is consumed before repainting. The first poll still blocks for 100 ms, so an idle TUI costs nothing. A burst is capped at 32 events (`EVENT_DRAIN_LIMIT`) so continuous input keeps showing progress instead of sitting on a stale frame.

Driving tuicr in a real pty, 20 keys with no pacing:

| | frames | bytes |
| --- | --- | --- |
| before | 22 | 30.8 KB |
| after | 1 | 20.7 KB |

The `while` condition also re-checks `app.should_quit`, so the Ctrl+C-twice path still exits from inside a burst rather than draining first.

`event_drain_timeout` is a small pure function specifically so the blocking-then-nonblocking-then-capped behaviour has a unit test without needing a terminal.

This explains the symptom ordering users report: wheel scrolling (many events per gesture) feels worse than held `j`/`k` (one per repeat), which feels worse than a single `m`/`M` jump.

## `test(perf): measure frame time against diff size and comment count`

Renders one frame with `TestBackend` against a synthetic diff at growing file counts and comment counts, and prints the median. `#[ignore]`d: `cargo test --release render_perf -- --ignored --nocapture`.

On current `main` it shows diff size is nearly free, ~1 ms per 40 000 diff lines:

```
   1 files x 200 lines =   200 diff lines: 0.68ms
 200 files x 200 lines = 40000 diff lines: 1.39ms
```

and comment count is not, at ~0.25 ms per body:

```
  20 files x 200 lines,   0 comments:  0.64ms
  20 files x 200 lines, 200 comments: 49.26ms
```

That second table is #552. @joshvito, feel free to lift this harness into #600 if it's useful for showing the effect of the cache there; with #600's fix applied locally the 200-comment row drops to 1.66 ms.

No timing assertions: they flake on shared CI, and the printed medians are what located the cost in the first place.

## Notes

- Deliberately not addressed: per-frame byte volume (~1.6 bytes/cell is near the floor, since per-row diff backgrounds mean a scroll changes every cell) and the O(total diff) line rebuild (~1 ms per 40 000 lines).
- Dogfooded: reviewed in tuicr, on the PR whose comment count started this.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,6 +33,22 @@ const CTRL_C_EXIT_TIMEOUT: Duration = Duration::from_secs(2);
 /// Hide the file list by default on narrow terminals.
 const MIN_WIDTH_FOR_FILE_LIST: u16 = 100;
 
+/// Upper bound on events consumed between two repaints. High enough that a whole
+/// trackpad gesture collapses into one frame, low enough that held-key
+/// auto-repeat still shows progress instead of sitting on a stale frame.
+const EVENT_DRAIN_LIMIT: usize = 32;
+
+/// How long the loop may wait for the next event after already consuming
+/// `drained` of them: the first blocks so an idle TUI costs nothing, later ones
+/// are only taken if already queued. `None` ends the burst.
+fn event_drain_timeout(drained: usize) -> Option<Duration> {
+    match drained {
+        0 => Some(Duration::from_millis(100)),
+        n if n < EVENT_DRAIN_LIMIT => Some(Duration::ZERO),
+        _ => None,
+    }
+}
+
 fn main() -> anyhow::Result<()> {
     profile::init_from_env();
 
@@ -415,8 +431,16 @@ fn main() -> anyhow::Result<()> {
             needs_redraw = false;
         }
 
-        // Handle events
-        if event::poll(Duration::from_millis(100))? {
+        // Handle events. Every already-queued event is consumed before the next
+        // repaint: a trackpad gesture or held key delivers events faster than a
+        // terminal presents frames, and the intermediate frames are never seen.
+        // `drained` caps a burst so continuous input still repaints.
+        let mut drained = 0;
+        while !app.should_quit
+            && let Some(timeout) = event_drain_timeout(drained)
+            && event::poll(timeout)?
+        {
+            drained += 1;
             // Set before reading so the many `continue` paths below (leader
             // keys, zz/ZZ, dd, {count}G, …) still schedule a redraw on the
             // next iteration even though they short-circuit past the match.
@@ -921,4 +945,28 @@ fn run_editor_from_tui<W: Write>(
     let editor_result = tuicr::editor::run_editor(&command);
     suspension.resume()?;
     Ok(editor_result.map(|()| EditorOutcome::Finished))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn drain_blocks_once_then_takes_only_queued_events() {
+        assert_eq!(
+            event_drain_timeout(0),
+            Some(Duration::from_millis(100)),
+            "the first poll must block, or an idle TUI spins"
+        );
+        assert_eq!(event_drain_timeout(1), Some(Duration::ZERO));
+        assert_eq!(
+            event_drain_timeout(EVENT_DRAIN_LIMIT - 1),
+            Some(Duration::ZERO)
+        );
+        assert_eq!(
+            event_drain_timeout(EVENT_DRAIN_LIMIT),
+            None,
+            "a capped burst must repaint instead of draining forever"
+        );
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,7 +43,7 @@ const EVENT_DRAIN_LIMIT: usize = 32;
 /// are only taken if already queued. `None` ends the burst.
 fn event_drain_timeout(drained: usize) -> Option<Duration> {
     match drained {
-        0 => Some(Duration::from_millis(100)),
+        0 => Some(Duration::from_millis(50)),
         n if n < EVENT_DRAIN_LIMIT => Some(Duration::ZERO),
         _ => None,
     }
--- a/src/app/tests/mod.rs
+++ b/src/app/tests/mod.rs
@@ -11,6 +11,7 @@ mod file_filter_tests;
 mod find_source_line_tests;
 mod persistence_merge_tests;
 mod pr_info_tests;
+mod render_perf_tests;
 mod scroll_behavior_tests;
 mod scroll_tests;
 mod single_file_view_tests;
--- a/src/app/tests/render_perf_tests.rs
+++ b/src/app/tests/render_perf_tests.rs
@@ -1,0 +1,230 @@
+//! Measures one `ui::render` frame against diffs of increasing size, so the
+//! per-frame cost can be compared across branches and checked for O(total
+//! diff) rather than O(visible rows) scaling.
+//!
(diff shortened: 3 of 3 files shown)


src/main.rs
```diff
@@ -33,6 +33,22 @@ const CTRL_C_EXIT_TIMEOUT: Duration = Duration::from_secs(2);
...
+/// trackpad gesture collapses into one frame, low enough that held-key
+/// auto-repeat still shows progress instead of sitting on a stale frame.
+const EVENT_DRAIN_LIMIT: usize = 32;
+
+/// How long the loop may wait for the next event after already consuming
+/// `drained` of them: the first blocks so an idle TUI costs nothing, later ones
+/// are only taken if already queued. `None` ends the burst.
+fn event_drain_timeout(drained: usize) -> Option<Duration> {
+    match drained {
+        0 => Some(Duration::from_millis(100)),
```
agavra: I think we should lower this to ~30ms, 100ms is like 10fps which is pretty slow for a TUI.

agavra (approved):
Love this improvement, makes a lot of sense but I think we should lower the coalescing treshold

joshvito:
@KallivdH thanks — took you up on the offer and lifted the harness into #600 (`cherry-pick -x` of ad5ae54, so it keeps your authorship): https://github.com/agavra/tuicr/pull/600/commits/16047e2904e0fdcfb87b767ee4d02c8541a2e0ea

It applied clean on top of the v0.22.0 merge, no adaptation needed.

Your 200-comment row on my hardware, measured both ways on the same machine — baseline is a worktree at ad5ae54 (main + harness, none of #600's changes), so the two rows are directly comparable:

**20 files x 200 lines, varying comment count**

| comments | main | #600 |
| --- | ---: | ---: |
| 0 | 0.62 ms | 0.59 ms |
| 20 | 7.16 ms | 0.70 ms |
| 60 | 18.24 ms | 0.65 ms |
| 200 | 57.65 ms | 0.84 ms |

68x at 200 comments, but the shape is the part that answers #552: the cost was linear in total comment volume and is now flat — 200 comments cost about what 0 do. That is the memoization plus the viewport culling; a body is highlighted once per session instead of once per comment per frame, and boxes off screen are not formatted at all.

Diff-size scaling is unchanged, as expected — nothing in #600 touches it:

| diff lines | main | #600 |
| --- | ---: | ---: |
| 200 | 0.42 ms | 0.50 ms |
| 40000 | 3.07 ms | 2.68 ms |

The harness is what made the flat line legible instead of "feels better", so it was worth having in-tree — thanks for writing it.

On the two PRs: they multiply rather than overlap. You cut how many frames get drawn, #600 cuts what each frame costs, and a wheel gesture over a comment-heavy PR pays both. No file overlap either (`main.rs` + the test module vs `syntax/mod.rs` + `ui/*`), so no reconciliation needed — the only shared file is the harness, and #600 carries your commit verbatim.

joshvito:
Correction to my numbers above: the regex-backend part of that comment was wrong, and I have reverted it out of #600.

I had benchmarked the backend swap (syntect `fancy-regex` instead of onig) on one markdown line packed with inline code spans, which is the only shape where fancy-regex wins. Measured across realistic bodies, same machine, per markdown-highlight call:

| markdown body | onig | fancy-regex |
| --- | ---: | ---: |
| body with a fenced ` ```rust ` block | 0.27 ms | 4.24 ms |
| plain prose | 0.04 ms | 0.13 ms |
| one line, 12 inline code spans | 52.17 ms | 17.75 ms |

A fenced block embeds a second grammar, which is exactly what `comment_body` in your harness generates — and what a real review comment carries. Running your harness against the backend swap on its own put 200 comments at **834.20 ms/frame**, against 56.22 ms on `main`. So the "onig costs 50 ms where fancy-regex costs 15 ms" line in my earlier comment was true only for that one synthetic line and misleading as a general claim.

Your harness is what caught it — the swap looked free until something measured a body shaped like a real comment.

#600 is now memoization plus viewport culling only, with `Cargo.toml` and `Cargo.lock` identical to `main`. Updated numbers there, on the current base:

| comments | main | #600 |
| ---: | ---: | ---: |
| 0 | 0.57 ms | 0.60 ms |
| 20 | 6.77 ms | 0.77 ms |
| 60 | 17.89 ms | 0.74 ms |
| 200 | 56.22 ms | 0.95 ms |

Issue #624, opened 2026-08-19
`e` opens the worktree copy in PR mode, not the PR revision — misleading "file does not exist"


joshvito:
tuicr 0.22.0 (debug build at `4b1be9a`), Windows 11, Azure DevOps PR mode.

### Repro

1. Repo has a PR whose head branch contains `some/new/file.txt`; check out an unrelated branch that does not contain that path.
2. Open the PR in tuicr, focus that file in the diff, press `e`.
3. → `Cannot open <checkout>/some/new/file.txt: file does not exist`

The file is real: it exists in the PR head commit and its blob is in the local object store (repo was fetched). It just isn't in the checked-out tree.

### Cause

`queue_editor_for_file_idx` (`src/app/reviewed.rs:138-178`) targets the worktree and requires the file on disk:

```rust
let path = root.join(&display_path);   // root = worktree / local_checkout_path()
if !path.exists() { /* "file does not exist" */ }
```

In PR mode the diff is the PR's revision, but the editor gets whatever the worktree holds. Two problems:

1. **Wrong diagnosis.** The blob is available locally at the PR head SHA — tuicr already reads it that way for context expansion (`read_blob_with_repo` → `git show <sha>:<path>`, `src/forge/azure/az.rs:243` and the gh/glab/bkt mirrors). "file does not exist" reads like tuicr lost the repo; the real cause is checkout revision ≠ PR head.
2. **Silent wrong revision (bigger).** When the path *does* exist in the worktree but at different content, `e` opens that revision and jumps to a line number derived from the PR diff — wrong text, wrong line, no warning. Only `path.exists()` is checked; nothing compares revisions. Hits every PR review where the PR branch isn't checked out.

### Possible fixes

1. In PR mode, when worktree HEAD ≠ PR head SHA, say so (name both) instead of "file does not exist".
2. Compare the worktree blob for the path against `<pr_head_sha>:<path>`; warn or refuse when they differ.
3. When the worktree copy is missing or differs, extract the PR-head blob to a temp file (reusing the existing `git show <sha>:<path>` path) and open it read-only, so content and line anchors match the diff.

joshvito:
What do we think the best path forward is? Fix 1 is clearly the shortest answer to better understanding the issue, and the user can just do the workaround checkout. That ignores the second issue though.

chrishenzie:
[...]

PR #627, opened 2026-08-19
fix(editor): open the PR revision with 'e', not the worktree copy

joshvito:
Draft for issue #624

In PR review, `e` resolved its target as <checkout>/<path> and only checked that the path existed, so it opened whatever revision the working tree happened to hold — silently, at a line number computed from the PR diff — or refused with "file does not exist" when the checkout lacked the file at all.

`e` now opens the reviewed revision. The worktree file is still preferred when its content matches, since it is the only copy edits can land in; otherwise the PR content goes to a read-only snapshot under the temp dir, opened at the same line. Files deleted by the PR open from the base side. Binary files are only ever opened from the checkout — blob and API reads are lossy UTF-8.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,7 @@ src/
 ├── config/
 │   └── mod.rs           # User config loading (XDG on Unix, %APPDATA% on Windows)
 ├── app.rs               # Application state (App struct, InputMode, etc.)
+│   ├── editor_target.rs # Read-only snapshots of a PR revision for `$EDITOR`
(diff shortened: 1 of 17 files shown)


chrishenzie:
Is anything blocking opening this for review?

joshvito:
Nope, was just using it as a daily driver, and wanted to find any issue's first.

agavra:
Gave it a quick test, LGTM. The only issue we can address in a follow-up is that it seems like it ignores the selected commit range. I think that's probably fine, but maybe we want to fail hard if you're trying to edit a non-head commit from a PR.

PR #630, opened 2026-08-20
feat(config): add confirm_quit to gate the bare q quit

aeslaughter:
I'm not a regular vi/vim user. I hit `q` on reflex — not a deliberate quit, just muscle memory from my emacs transient setup — and it closed the review I'd been building, which caught me off guard. I'd like a way to make quitting something I opt into, rather than one keystroke away.

Thanks for tuicr — it's genuinely helpful, and this is the only rough edge I've hit.

The guard meant to catch this looks unreachable in practice: it needs `dirty && session.has_comments()` in `handle_shared_normal_action`, but adding a comment autosaves and clears `dirty`, so one `q` still exits. vim diverges here too — normal-mode `q` records a macro, and only `:q` quits. Refs #427, where someone hit the same surprise from another angle.

`confirm_quit` defaults to `false`, which is today's behavior unchanged. Set it to `true` and bare `q` asks first.

The gate sits on the single `Action::Quit` arm in
`handle_shared_normal_action`, so no explicit quit path can reach it: `:q`/`:q!`/`:wq`/`:x` dispatch through `CommandKind`, and `ZZ`/`ZQ` are handled inline in the main loop. The `q` bindings for the help overlay, summary view, and review target selector have their own actions and are untouched.

`ConfirmAction` now carries its own prompt, since a quit prompt asks whether to leave at all — "no" returns to the review, unlike `CopyAndQuit`, which only asks whether to export on the way out.

One pre-existing fix, found by dogfooding this change: the confirm dialog was sized `centered_rect(50, 20)`, a share of the terminal, which left a short prompt floating in a huge mostly-empty box on a large screen. It is now sized to its contents. This also tightens the existing clipboard prompt — happy to split it out if you'd rather keep this PR to the config key.

Written with claude-code. I tested and reviewed it myself, including reviewing this diff in tuicr per CONTRIBUTING — which is how the oversized confirm dialog turned up.

### DIFF
--- a/src/input/keybindings.rs
+++ b/src/input/keybindings.rs
@@ -55,6 +55,10 @@ pub enum Action {
 
     // Session
     Quit,
+    /// Transitional hint for the removed `q`-quits binding: shows a message
+    /// pointing at `:q` instead of quitting. Drop this a few releases after
+    /// the `q` removal has had time to land.
+    QuitHint,
     ExportToClipboard,
     /// Copy just the comment under the cursor (`Y`), not the whole review.
     CopyCommentAtCursor,
@@ -238,6 +242,9 @@ fn map_normal_mode(key: KeyEvent, leader_key: char) -> Action {
         (KeyCode::Char('/'), _) => Action::EnterSearchMode,
         (KeyCode::Char('?'), _) => Action::ToggleHelp,
         (KeyCode::Esc, KeyModifiers::NONE) => Action::ClearSearchHighlight,
+
+        // Transitional hint: q used to quit; now it just points at :q.
+        (KeyCode::Char('q'), KeyModifiers::NONE) => Action::QuitHint,
 
         (KeyCode::Char(' '), KeyModifiers::NONE) => Action::ToggleExpand,
         (KeyCode::Char('o'), KeyModifiers::NONE) => Action::ExpandAll,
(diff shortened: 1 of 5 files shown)


agavra:
I'm actually happy to remove `q` to quit and stick to vim defaults where only `:q` quits (or `:x` etc...). What do you think? That was somewhat of a legacy choice early on.

aeslaughter:
Removing is fine with me, simpler is always better. Thanks!

src/input/keybindings.rs
```diff
@@ -517,7 +512,6 @@ fn map_visual_mode(key: KeyEvent) -> Action {
         (KeyCode::Char('y'), KeyModifiers::NONE) => Action::ExportToClipboard,
         (KeyCode::Esc, KeyModifiers::NONE) => Action::ExitMode,
         (KeyCode::Char('v') | KeyCode::Char('V'), _) => Action::ExitMode,
-        (KeyCode::Char('q'), KeyModifiers::NONE) => Action::Quit,
```
agavra, line 520: sorry for the back and forth on this, but I just thought it might be worth for a few versions at least to display a message (bottom right as with existing messages) that says "use :q to quit" or something like that so people that relied on `q` aren't surprised by the backwards incompatibility

agavra (approved):

PR #634, opened 2026-08-20
side-by-side: choose the comment side with ←/→, caret follows

thiblahute:
In side-by-side view the side a new comment attaches to was auto-derived per
line (new preferred), and the `▶` caret sat at column 0 — so a line with both
sides could only be commented on the new one, and nothing on screen said which
side `c` would pick.

Adds an active `cursor_side` (default `New`). `get_line_at_cursor` resolves it,
clamped to the sides a line actually has, and a post-render overlay moves the
`▶` caret onto the active side's gutter.

The two columns are stops on the existing `<leader>h` / `<leader>l` panel walk
— file list ← old ← new — so arrows and `h`/`l` keep scrolling everywhere, as
today. Entering the diff from the file list keeps whichever side was active;
unified view is unchanged.

Tested with `src/app/tests/sbs_comment_side_tests.rs`; `cargo fmt --all
--check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` (1534
passed) are green.

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -109,8 +109,8 @@ considers files that pass the active filters.
 | Key | Action |
 |-----|--------|
 | `Tab` / `Shift-Tab` | Cycle focus forward / backward between file list, comment navigator, diff, and commit selector |
-| `<leader>h` | Focus file list (left panel) |
-| `<leader>l` | Focus diff view (right panel) |
+| `<leader>h` | Move focus one panel left (side-by-side: new side → old side → file list) |
+| `<leader>l` | Move focus one panel right (side-by-side: file list → diff → old side → new side) |
 | `<leader>k` | Move focus up (comments to files, or diff/files to commit selector when visible) |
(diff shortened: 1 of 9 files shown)


agavra:
I like this feature but I'm worried about the key bindings, I think we probably want to make side by side diff behave like full panels instead of psuedo-panels and then we use `<leader>+l` and `<leader>+h` to move betewwen them. Right now we don't have a way to "scroll" the side by side diff, which we really should add. Once we do, I'd expect the arrows and `h` and `l` to scroll not switch panels.

thiblahute:
Right, redone. Arrows and h/l scroll again, sides are stops on the
`<leader>h`/`<leader>l` walk instead. Didn't make them real panels internally,
say if you want that.

agavra:
This is fine for now, thanks! I'll take a look soon.

agavra:
Ah I think you meant to push the change?

thiblahute:
Indeed, done.

PR #635, opened 2026-08-20
docs: add pacman installation

tridimensionaal:
im relatively new using tuicr, and while checking how to install it i found that it is also available in the official arch linux repositories and can be installed with pacman

it seems like a good idea to add that option to the README so other Arch users can find it easily

package: [arch linux `extra/tuicr`](https://archlinux.org/packages/extra/x86_64/tuicr/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@
 curl -fsSL tuicr.dev/install.sh | sh
 # or
 brew install tuicr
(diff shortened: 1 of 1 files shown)


agavra (approved):
Nice, I had no idea it was possible to install it this way!

PR #642, opened 2026-08-21
vcs/jj: use as_chunks to parse diff metadata

samuelkarp:
Replace `chunks_exact(3)` with `as_chunks::<3>()` when parsing Jujutsu null-delimited diff metadata records, addressing the Rust 1.98.0 `clippy::chunks_exact_to_as_chunks` lint.

Assisted-by: Antigravity


agavra (approved):
thanks! not sure what broke clippy

samuelkarp:
It was a new check for Rust 1.98 which just came out.

PR #646, opened 2026-08-25
fix(ui): render a visible stand-in for whitespace leader keys in help

OddMagnet:
Hey, first of all, great project! Hope you don't mind a small feature/fix pr.

**Symptom**

`leader` accepts any single character (`read_leader`, `src/config/mod.rs`), so `leader = " "` is a valid config - and it works correctly at the keymap level, space really does trigger the chords. 
The help popup (?), however, interpolated the raw char into its lines, so every leader chord rendered as a blank gap:

```
Tab/S-Tab Toggle focus next/previous panel
h/ l      Focus file list/diff
e         Toggle file list visibility
s         Toggle commit selector visibility
```

**On the column width**

The help lines pad each chord to a fixed width (`"  {leader}h/{leader}l     "`), so the substitute must occupy exactly one display column. a word like "Space" would shift the whole help block out of alignment. 
The helper returns `char` rather than `String` to make that structural, and a test asserts the substitute's display width is 1.
␣ (U+2423) is narrow in both `width` and `width_cjk`, so those columns hold everywhere. 

**Tests**

Adds a `mod tests` to `help_popup.rs`: whitespace leader -> stand-in, ordinary leader -> unchanged, and stand-in is one display column wide.

`cargo test` (1528 passed), `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --all --check` are clean. 

Verified by hand in the TUI with both leader = " " and leader = ";".

No docs changed since this is neither a new keybinding nor a new config key.

### DIFF
--- a/src/ui/help_popup.rs
+++ b/src/ui/help_popup.rs
@@ -62,8 +62,22 @@ pub fn render_message_details(frame: &mut Frame, app: &mut App) {
     );
 }
 
+/// Printable stand-in for the configured leader key.
+///
+/// A whitespace leader (`leader = " "`) is a valid config value but renders as a
+/// blank in the help text, so substitute a visible glyph. The substitute must be
+/// exactly one display column wide: the help lines pad each chord to a fixed width.
+fn leader_display(leader: char) -> char {
+    if leader.is_whitespace() {
+        '\u{2423}'
+    } else {
+        leader
+    }
+}
+
 pub fn render_help(frame: &mut Frame, app: &mut App) {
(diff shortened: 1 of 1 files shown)


OddMagnet:
Not sure why Clippy failed, anything I can/need to do about it?

agavra:
Clippy failed for something unrelated due to rust toolchain upgrade. Taking a look now!

agavra (approved):

agavra:
Thanks for the contribution @OddMagnet!

PR #648, opened 2026-08-25
fix(tui): resume from an editor without querying the cursor

r-vdp:
Closing a full-screen editor could leave tuicr showing "Failed to
restore terminal: The cursor position could not be read within a
normal duration" over a screen that never repainted.

`TerminalSuspension::resume` called `Terminal::clear`, which starts by
reading the cursor position — an `ESC [ 6 n` query written to
`io::stdout()`, not to the writer the TUI draws on. When stdout is not
the terminal (`--stdout` redirects it to a file) no reply can arrive,
so the two-second timeout fires every time, aborting the clear before
it resets the back buffer. The next draw diffs against the pre-editor
screen and paints nothing.

Resize to the current size instead: the same clear and reset, but
sized from `/dev/tty`.

### DIFF
--- a/src/terminal_state.rs
+++ b/src/terminal_state.rs
@@ -134,6 +134,23 @@ impl<W: Write> TerminalSession<W> {
         self.active = false;
         Ok(())
     }
+
+    /// Clears the screen and drops the back buffer so the next draw repaints
+    /// every cell.
+    ///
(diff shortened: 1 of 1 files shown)


r-vdp:
Clippy error seems unrelated.

agavra (approved):
Thanks! I'll fix the clippy error on main, I wonder what caused it.. maybe update in rust versions

PR #651, opened 2026-08-26
fix: Make commit branch columns responsive

### LINKED ISSUE #618: Branch name column in commits view is too narrow
Hi,

I am mostly using tuicr for final review of my own PRs before submitting a branch as PR.

The branch column is really narrow and does not distinguish branches if you work with prefixes:

Image

Is there a way to make it bigger? Why isn't more screen space used?

Cheers
joe

mvanhorn:
Commit rows always reserve 16 cells for the branch chip and 50 cells for the summary, regardless of the terminal width. Long prefixed branch names therefore collapse to indistinguishable chips even when the selector has substantial unused horizontal space. The shared row renderer affects both the startup review-target selector and the inline commit selector, so the correction must stay consistent across both surfaces. The owner has confirmed that making this layout responsive is the intended behavior.

## Summary

Extend the shared commit-row rendering input in `src/ui/commit_row.rs` with the content width available to that row, then derive aligned branch and summary column widths from a saturating horizontal budget after reserving the controls, hash, and author/date metadata. Preserve sensible minimum widths around the current compact layout, but distribute wider-screen surplus between branch and summary with the summary retaining the larger share; this reveals materially more of prefixed branch names without sacrificing the commit-message-first design. Keep the calculation in the production row renderer and use one width for every row so commits without a branch still pad to the same summary start.

## Validation

- Render a wide commit row containing two long branch names with the same prefix and verify the responsive branch column exposes distinguishing suffix characters while the summary and author/date remain present.
- Render at the existing compact-width budget and verify branch, summary, selection/review markers, hash, author, and relative date retain a stable aligned layout without arithmetic underflow or overflow.
- Render otherwise identical rows with and without a branch and verify their summaries begin in the same column for a given available width.
- Render a very narrow row and verify saturating width allocation produces a valid truncated line rather than panicking; the enclosing `Paragraph` may clip metadata that cannot fit.
- Render staged and unstaged pseudo-commit rows at multiple widths and verify their special tags and summaries remain unchanged by normal commit-column allocation.
- Draw both the fullscreen local target selector and bordered inline commit selector and verify each passes its actual usable width, so resizing either surface changes branch visibility consistently and borders are not counted as content space.

Fixes #618

AI was used for assistance.

### DIFF
--- a/src/ui/commit_row.rs
+++ b/src/ui/commit_row.rs
@@ -20,15 +20,20 @@ pub const UNSELECTED_BOX_GLYPH: &str = "\u{25a2}"; // ▢
 pub const REVIEWED_GLYPH: &str = "\u{2713}"; // ✓
 pub const REVIEWED_LABEL: &str = "✓ ";
 
-// Fixed column widths so author/date land at the same x across every row.
-// Branch column gets `[branch_name]` padded to width including brackets and a
-// trailing space; rows without a branch render the same number of blanks.
-const BRANCH_COL_WIDTH: usize = 16;
-const SUMMARY_COL_WIDTH: usize = 50;
+// Baseline column widths preserve the compact layout. Narrow rows shrink both
+// columns proportionally; wider rows give one third of their surplus to the
+// branch and two thirds to the commit summary.
+const COMPACT_BRANCH_COL_WIDTH: usize = 16;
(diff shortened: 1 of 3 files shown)


agavra (approved):
Thanks @mvanhorn :)

Issue #518, opened 2026-07-30
Support Forgejo as review target


dwt:
I find that I use Forgejo more and more besides the two large ones, so support would be much appreciated.

drjackyl:
(Including self hosted Forgejo)

agavra:
Makes sense to me! I don't have a forgejo account/repo to host against. Bitbucket is being added and adding new forges is pretty simple, feel free to submit a patch yourself and I can take a quick look.

PR #652, opened 2026-08-26
feat(forge): add Gitea pull request review support

ghaith:
## Summary

Adds Gitea pull request review through the [`tea`](https://gitea.com/gitea/tea) CLI, in `src/forge/gitea/` alongside the GitLab, Bitbucket, and Azure backends. Relates to #518 and could be ground work for it but does not solve it. See the note on #615 at the bottom.

The full `ForgeBackend` surface is implemented, nothing stubbed: PR listing (open and review-requested), PR open, cumulative and commit-range diffs, commit lists, review threads and summaries, remote context expansion, the description panel, and `create_review` with Comment, Approve, Request changes, and Draft. A Gitea review is one `POST`, so inline comments and the summary land atomically.

## Scope: Gitea only

Forgejo and Codeberg are **not** claimed. They currently serve a compatible API and will route here if the user has a `tea` login for them, but their hostnames are deliberately not recognized — they are separate projects free to diverge, and this is tested against neither. `docs/GITEA.md` says so explicitly, including "don't file fork breakage as a Gitea bug".

## Three Gitea/`tea` behaviors worth knowing

**`tea api` exits 0 for every HTTP response.** It writes the body to stdout whether the status is 200 or 404, so a non-zero exit never arrives and `run_command_output` hands back an error page as if it were data. The backend passes `-i`, which puts the status line on stderr, and reads it from there. That is what the new `run_command_streams` in `src/process.rs` is for — the existing helper collapses a non-zero exit into an error and discards stderr on success, and neither applies here.

**Gitea clamps `limit` to `MaxResponseItems`** (default 50, instance-configurable). So "stop when fewer rows come back than were requested" truncates at the first page. Pagination drives off `X-Total-Count`, falling back to comparing against the *first page's* size rather than the requested size. This matters most for `/pulls/{n}/files`, which is paired positionally with the `.diff` text: a short read is a hard `metadata records but N patch blocks` mismatch, not a silent gap.

**`/pulls/{n}/commits` returns commits newest-first**, the reverse of the order the trait specifies. Left as-is it inverts the commit selector and makes commit-range scoping pass the later commit as `start_sha`, which `/compare` answers with an empty diff.

Also: Gitea reports a mode-only change as file status `unchanged` while still emitting a `diff --git` block, so those files have to stay in the metadata list; and it spells review events `APPROVED`/`REQUEST_CHANGES` where GitHub says `APPROVE`/`CHANGES_REQUESTED`. All of the above are captured as gotchas 17–19 in `AGENTS.md`.

## Verification

47 unit tests against a fake runner, plus an end-to-end pass against **Gitea 1.27.2 in Docker**: a PR covering add, modify, delete, rename, and a mode-only change, with review threads on both sides, context expansion, a pending Draft, and a submitted Comment review read back to confirm inline comments landed on the requested lines.

Two of the three behaviors above were found *by* that live pass — the commit ordering and the pagination cap are invisible to a fake runner.

Separately, driven by hand through the TUI: commenting and approving on a local instance, PR listing against a real self-hosted Gitea, the `tuicr pr <url>` target form, and a check that GitHub repositories still resolve as before.

`cargo fmt --check` clean, `cargo clippy --all-targets` no warnings, 1574 tests pass.

On dogfooding: the Gitea path itself was exercised by reviewing and submitting through tuicr against the live instance. This diff was reviewed in tuicr by hand.

## Known limitations

All documented in `docs/GITEA.md`:

- **Multi-line comments collapse to their last line.** Gitea's review API takes one line number per comment and has no range form. Comment text is never altered.
- **No thread is marked outdated.** Gitea tracks force-push invalidation internally but does not expose it on the API.
- **Review-requested filtering is approximate.** Gitea exposes that filter only on a cross-repository issue search, so those rows carry no branch names and the selector's branch filter does not match them.
- **Commit-range diffs on git-quoted filenames need a local clone.** `/compare` returns a raw diff with no structured file list, so paths are read from the diff headers; git quotes paths with control or non-ASCII bytes, which this does not unquote. The local-clone fast path covers it.

## Relationship to #615

#615 (@kennethdsheridan, draft) got to this ground first, for Forgejo and Codeberg. It doesn't mention Gitea, but the two patches land in the same place, and I'd rather flag that up front than have it surface as a merge conflict.

Detection is the reason. Both patches recognize a forge by "the host has a `tea` login" and by the `/pulls/<n>` URL shape. `tea` is Gitea's CLI, and a tea login says nothing about which fork is behind a hostname — so there is one tea-shaped slot in the detection chain and both patches fill it. In #615 that means a plain Gitea instance detects as `ForgeKind::Forgejo`; here a Forgejo instance detects as `ForgeKind::Gitea`. Neither is avoidable, and the two can't coexist: whichever parser runs first in `parse_any_remote_url` wins and the other becomes unreachable.

Three concrete overlaps:

1. Same source lines — the `ForgeKind` enum, both `slug.rs` arms, the `comment_panel` badge, `create_forge_backend`, `parse_any_remote_url`, `parse_url_target`. Mechanical.
2. The unreachable-parser problem above.
3. **The persisted slug prefix**, `gt:` vs `fj:`. `slug_for_session()`'s string is hashed into the on-disk session path and used as the manifest key, so it is session identity. If one ships and users accumulate sessions under its prefix, a later rename orphans their comments and reviewed-markers. This is the one worth getting right the first time.

The scopes also differ: #615 covers discovery — list, open, diff — and fails closed on submit, commits, range diffs, threads, and context expansion; this covers the whole trait.

My suggestion, entirely yours to overrule: land the Gitea backend under the name of the thing that's actually tested, and let #615 become the fork-specific work — Forgejo and Codeberg handling, on top of this, once and where they genuinely diverge. That seems like the better split of effort than two implementations of the same API, and it avoids naming a backend after a fork it hasn't been tested against.

I'm happy to rebase onto #615 instead, fold this into it, or change the kind name and slug prefix — whichever you prefer. Probably easiest to settle during review rather than guessing now.


kennethdsheridan:
Thanks for calling out the overlap with #615. I agree with the direction of landing this as Gitea support under a Gitea kind and slug, and treating Forgejo and Codeberg as separate follow up work.

The one thing I would push back on is the closing language for #518. #518 is specifically asking for Forgejo, including self hosted Forgejo, and this PR is deliberately scoped and tested as Gitea. I think this should reference #518 as related or as unblocking adjacent forge work, but not close it.

The persisted slug prefix is the key point for me. gt should mean Gitea, and any future Forgejo work should get its own stable fj identity from the start. I will rework or close #615 accordingly rather than trying to make two tea based parsers compete.

agavra (approved):
Thank you Ghaith for the contribution! This makes sense to me.

Issue #639, opened 2026-08-21
review list never surfaces PR sessions in an Azure DevOps checkout (owner resolves to _git)


joshvito:
`tuicr review list` in a checkout is documented to show that checkout's local sessions *plus* the PR sessions for the same repo (`docs/REVIEW_CLI.md`). In an Azure DevOps checkout it only shows the local sessions — the repo's PR sessions are invisible unless you pass `--all`.

### Repro

1. Clone an Azure DevOps repo (`origin` = `https://myorg@dev.azure.com/myorg/myproject/_git/myrepo`).
2. Open a PR review and a local review in that checkout so both sessions persist.
3. Run `tuicr review list` from the checkout.

**Expected:** both sessions.
**Actual:** only the local one. `tuicr review list --all` shows the PR session as `az:myorg/myproject/myrepo/pr/123`.

### Cause

`slug::parse_remote_owner_repo` / `parse_path_segments` (`src/slug.rs`) always take the last two path segments of the remote URL. Azure URLs are `host/org/project/_git/repo`, so the owner comes out as `_git`. PR sessions get their owner from the forge layer, which correctly packs `org/project` (`ForgeRepository::azure`). `RepoCoordinate::matches` then compares `"_git"` against `"myorg/myproject"` and rejects it, so the PR entry is filtered out.

Same root cause, a few more symptoms:

- Local Azure slugs read `_git/myrepo@branch/...` instead of `myorg/myproject/myrepo@branch/...`.
- `tuicr review list --repo dev.azure.com/myorg/myproject/_git/myrepo` (and the `forge:` form) fails for the same reason — `RepoCoordinate::parse` uses the same last-two-segments rule.
- SSH Azure remotes (`git@ssh.dev.azure.com:v3/org/project/repo`) resolve to owner `project`, also wrong.

It ironically works if the checkout has *no* parseable origin: the owner is then `None` and the owner comparison is skipped.

This also breaks the agent workflow in `skills/tuicr/SKILL.md`, which tells agents to attach to the active session via `tuicr review list`.

### Suggested fix

The repo already has a correct Azure parser: `forge::parse_any_remote_url` (`src/forge/mod.rs`) delegates to `parse_azure_remote_url` and produces `owner = "org/project"`. `src/slug.rs` could use it (falling back to the current generic parser) in both `resolve_owner_repo` and `RepoCoordinate::parse`, rather than maintaining a second, disagreeing remote parser.

Two things to handle if the local slug string changes as a result:

[...]

PR #656, opened 2026-08-26
fix: resolve nested GitLab group owners as the full namespace

joshvito:
The GitLab half of the same bug class as #639, split out as suggested in #654 and rebased onto `main` now that #654 has landed.

`tuicr review list` in a nested-group checkout now surfaces that repo's PR sessions, not just the local ones, and `--repo gitlab.com/org/team/svc` matches them too.

A repo under `org/team` resolved to owner `team`: a checkout's owner came from the generic last-two-path-segments rule, while its PR sessions carry the whole namespace from the forge layer. `RepoCoordinate::matches` compares owners as one opaque string, so those PR entries were filtered out.

The fix adds GitLab to `parse_any_remote_url_by_hostname` as a hostname-only variant. The full `parse_gitlab_remote_url` also recognizes self-hosted instances on custom domains, but does so by spawning `glab config get host` and reading `~/.ssh/config`, and `resolve_owner_repo` runs on every session save. Those instances are therefore still missed and keep falling back to the generic rule.

Risk: local slugs in a subgroup checkout change shape (`team/svc@…` → `org/team/svc@…`), so existing sessions there won't be reattached by the TUI. They stay listed and readable by their old slug.

I have no GitLab instance to test against. The behavior is covered by unit tests at both layers, but a check against a real subgroup checkout would be welcome.

### DIFF
--- a/docs/REVIEW_CLI.md
+++ b/docs/REVIEW_CLI.md
@@ -67,6 +67,15 @@ tuicr review list --repo git@ssh.dev.azure.com:v3/myorg/myproject/myrepo
 #   -> [ ..., { "slug": "az:myorg/myproject/myrepo/pr/123", "kind": "pr", ... } ]
 ```
 
+GitLab groups nest the same way, so a repo under `group/subgroup` has
+`group/subgroup` as its owner:
+
+```bash
+tuicr review list --repo https://gitlab.com/org/team/svc
+tuicr review list --repo git@gitlab.com:org/team/svc.git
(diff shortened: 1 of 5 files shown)


joshvito:
Would be cool, if any gitlab users could check this branch before we merge it.

agavra:
I'm not a gitlab user either 😅 I'll wait to see if anyone who has this issue chimes in.

joshvito:
well, I got as far as creating a gitlab account over the weekend. I could do a quick test, but still would be useful for a daily user to test drive. I'll post the results of my smoke test once done.

PR #662, opened 2026-08-28
Pass through scope args to all wrappers

joshgummersall:
Quick one — this extends the `-- <scope>` pass-through pattern from the cmux wrapper to the tmux, Zellij, and Herdr wrappers too. Previously only cmux let you scope a review with `-- -w` or `-- -r <revset>`; the other three only took a directory arg, so the agent had to launch tuicr first and leave the user to pick staged/unstaged/commit-range by hand in the TUI.

What changed:

- Added a shared helper for parsing and formatting tuicr args, fixing up herdr, tmux, and zellij usages
- Updated SKILL.md to say "every wrapper" instead of calling out cmux specifically, and made passing a scope explicit/required instead of optional

Should be no observable changes for existing cmux usage — this just closes the gap for the other three wrappers.

Tested out in cmux, herdr, tmux, and zellij.

### DIFF
--- a/skills/tuicr/SKILL.md
+++ b/skills/tuicr/SKILL.md
@@ -72,11 +72,15 @@ When the user needs an interactive tuicr pane and no active session exists:
 
 | Environment | Action |
 |-------------|--------|
-| `$CMUX_WORKSPACE_ID` is set | Run `tuicr-wrapper-cmux.sh /path/to/repo` |
-| `$TMUX` is set | Run `tuicr-wrapper.sh /path/to/repo` |
-| `$ZELLIJ` is set | Run `tuicr-wrapper-zellij.sh /path/to/repo` |
-| `$HERDR_ENV` is `1` | Run `tuicr-wrapper-herdr.sh /path/to/repo` |
+| `$CMUX_WORKSPACE_ID` is set | Run `tuicr-wrapper-cmux.sh /path/to/repo -- <scope>` |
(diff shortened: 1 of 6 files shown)


joshgummersall:
Note: planning to submit a PR next that adds a summary to output that looks like:

```
❯ tuicr
tuicr-session: joshgummersall/tuicr@print-review-summary/worktree/f149734
tuicr-summary: reviewed 4/4 files, 0 comments added
```

so the agent knows I've actually reviewed files even if I don't have any comments. This fixes a recurring issue I've had where the agent isn't sure if the review closed unexpectedly.

That diff is here: https://github.com/joshgummersall/tuicr/compare/pass-through-scope-args-to-all-wrappers...joshgummersall:tuicr:print-review-summary?expand=1

joshgummersall:
Follow-up in ce0fd22: `tuicr-wrapper-herdr.sh` was the only wrapper that never passed `--stdout`, so a human reviewing through the Herdr pane still had to answer tuicr's save & copy confirm dialog on exit (tmux/zellij already avoided this). Gave Herdr the same `--stdout`-capture-and-print behavior, and hoisted the now-duplicated detection/reporting logic (`tuicr_stdout_supported`, `tuicr_report_stdout_output`) into `_tuicr-common.sh` so all three wrappers share one implementation.

agavra (approved):
makes sense, thanks @joshgummersall :)

joshgummersall:
> makes sense, thanks @joshgummersall :)

thank you for the amazing tool!

PR #663, opened 2026-08-29
chore: bump nix flake input (naersk) due to changes to crate API

a-h:
crates.io started returning 403 for crates.io/api/v1/crates/<name>/<version>/download when the request carries curl's default User-Agent — which is what Nix's fetchurl sends. Browser UA gets 200, so it's UA-based blocking.

naersk fetched crates from that endpoint. It was fixed in nix-community/naersk#391 (merged 2026-06-08) by switching to static.crates.io, see https://github.com/nix-community/naersk/issues/390 for the issue.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -30,11 +30,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1769799857,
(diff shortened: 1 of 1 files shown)


a-h:
Oops. I see this is a dupe of #661

PR #666, opened 2026-08-31
diff: drop the frame and fold the header when the diff is the sole pane

thiblahute:
When neither the file list nor the inline commit selector is shown, the diff's
border box is redundant chrome. Render it borderless so the diff uses the full
width, and fold the file name (after the brand) and the diff stats (after the
source cluster) into the single top header line — the reserved title row read
as a second header under the global one.

sole-pane-before-after

Framed layouts (any other pane visible) keep the file name in the frame title
as before. No new config key or keybinding: `App::is_diff_sole_pane()` derives
it from `show_file_list` and the inline commit selector.

`cargo fmt --all --check`, `cargo check`, `cargo clippy -- -D warnings` and
`cargo test` (1555 passed) are green locally.

### DIFF
--- a/src/app/commits.rs
+++ b/src/app/commits.rs
@@ -324,6 +324,13 @@ impl App {
         self.set_message(format!("Commit selector: {status}"));
     }
 
+    /// Whether the diff is the only visible pane — no file list (which also
+    /// carries the comment navigator) and no inline commit selector. Used to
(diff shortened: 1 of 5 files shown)


agavra (approved):
Thank you @thiblahute for the contribution! This looks good to me.

PR #668, opened 2026-09-01
fix: apply .tuicrignore before parsing PR diffs

### LINKED ISSUE #650: fix: .tuicrignore not applied in PR mode (hangs on large minified files)
## What happens

`.tuicrignore` is not applied in PR mode. On a PR that touches large
minified files (e.g. committed build bundles), `tuicr pr <n>` hangs for
45s+ and never opens, even with the offending path in `.tuicrignore`.

The docs say `.tuicrignore` "excludes matching files from all review
diffs" — PR mode is the exception.

## Why

PR mode builds the full syntax-highlighted, intra-line diff before any
ignore filter runs. The profiler span `startup.syntax_highlighter`
dominates. `-p <path>` has the same problem — it filters too late.

Local mode is fine: `tuicr -r main..HEAD` with the same `.tuicrignore`
(or a git `-diff` attribute) opens instantly.

## Repro

Any PR whose diff includes a large minified file (one ~30KB+ single
line). With that path in `.tuicrignore`:

```sh
printf 'dist/\n' > .tuicrignore
tuicr pr <n>            # hangs on the minified dist/*.js
tuicr -r main..HEAD     # instant, dist excluded
```

## Expected

`.tuicrignore` (and `-p`) exclude matching files before highlighting and
diffing in PR mode, as in local mode.

## Version

tuicr 0.24.0, macOS arm64, libgit2 backend

atirna:
PR mode applied `.tuicrignore` after `parse_file_patches`, so a diff that was configured to be excluded was still fully parsed and syntax-highlighted first — on a PR touching a large minified bundle, `tuicr pr <n>` spent its whole startup in the highlighter (the `startup.syntax_highlighter` span) and then dropped the file anyway. The filter now runs on the raw `FilePatch` list before the parser (`tuicrignore::filter_file_patches`, mirroring the existing `filter_diff_files`), so an ignored file never reaches the parser or the highlighter at all. Both PR-mode parse sites get the reorder: the initial open/reload path (`prepare_open_pr`) and the commit-subrange reload (`finish_pr_range_reload`).

Two consequences worth naming:

- An ignored patch can no longer fail the PR open on content the parser would reject — there's a test pinning exactly that (`should_drop_ignored_patches_before_parsing_them`, which fails on the old ordering).
- A PR whose every file is ignored now opens as an empty review instead of colliding with the "no file changes" error path (a PR that genuinely has zero patches from the forge still errors).

Behavior outside a checkout is unchanged: with no `local_checkout` there is still no filtering, matching the module's documented invariant.

Local mode was already fine — the libgit2 backend builds diffs from structured deltas rather than reparsing patch text, which is why `-r main..HEAD` was instant for the same `.tuicrignore`.

Fixes #650

### DIFF
--- a/src/app/pr.rs
+++ b/src/app/pr.rs
@@ -413,22 +413,23 @@ impl App {
         use crate::vcs::diff_parser::parse_file_patches;
 
         let highlighter = self.theme.syntax_highlighter();
+        let local_checkout = self
+            .forge_backend
+            .as_deref()
+            .and_then(|b| b.local_checkout_path());
+        // Filter before parsing so an ignored large file is never
+        // highlighted, exactly as in `prepare_open_pr`.
+        let patches = match local_checkout.as_deref() {
+            Some(root) => crate::tuicrignore::filter_file_patches(root, patches),
+            None => patches,
+        };
         let parsed = match parse_file_patches(patches, highlighter) {
             Ok(files) => files,
             Err(TuicrError::NoChanges) => Vec::new(),
             Err(e) => return Err(e),
         };
 
-        let local_checkout = self
-            .forge_backend
-            .as_deref()
-            .and_then(|b| b.local_checkout_path());
-        let files = match local_checkout.as_deref() {
-            Some(root) => crate::tuicrignore::filter_diff_files(root, parsed),
-            None => parsed,
-        };
-
-        self.diff_files = files;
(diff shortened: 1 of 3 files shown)


agavra (approved):
Thanks @atirna - this is a good optimization!

PR #669, opened 2026-09-01
feat(selector): add a Sessions tab to resume saved reviews

martintrojer:
Reopening a saved review meant remembering the `-r` range or `-w` flag it was created with. Get it wrong and tuicr silently starts a fresh session instead of the one holding your comments.

This adds a **Sessions** tab to the review target selector, listing saved reviews for the current checkout. Enter resumes the row with its stored target.

Sessions tab

`:sessions` (alias `:reviews`) opens the selector on the tab; Tab cycles three ways.

Notes:

- Commit ranges resume in their stored oldest-first order. Reversing it inverts the diff and creates a second session for the same review, with no error.
- `WorkingTree` sessions get their own loader — `SessionDiffSource` is part of a session's identity.
- Resume installs the session that was selected. The loaders each resolve one from the current branch and HEAD, which is not necessarily the row picked.
- Sessions with no comments and no reviewed files are hidden; they hold no progress to resume.
- A range whose commits no longer resolve is refused rather than partially loaded.

Tested with `cargo test --lib` (1570 passed) plus 18 new tests driven through `sessions_tab_select`. Dogfooded: this PR's own review was resumed from the tab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,11 @@ so collaborative tools can add comments immediately. Empty auto-created session
 files are removed when the TUI exits. `tuicr review list` marks currently open
 TUI sessions with `"active": true`.
 
+Inside the TUI, the review target selector's **Sessions** tab lists the saved
+reviews for the current checkout, so you can resume one by picking it instead of
+retyping the commit range it was opened with. Open it with `:sessions` or cycle
+to it with `Tab`.
+
 ## Library API
 
 tuicr also exposes a Rust library API for tools that want to build on top of its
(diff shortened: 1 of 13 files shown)


agavra:
This is SUCH a good idea @martintrojer 🙏 I tested it out and it seems to work nicely. The experience of resuming PR sessions is weird though, I get an error saying it has to be resumed from the PR tab. I think it'd be good if it allowed me to resume it anyway. If that's complicated for whatever reason it can be done as a follow up.

Another follow up that would be nice is the ability to delete a session from that tab as well using `dd` like deleting comments (though I think we should have a confirm dialog for that).

> Sessions with no comments and no reviewed files are hidden;

I think it'd be nice to "unhide" these with a keystroke so that you can then go through using `dd` and delete them. In general I see this as a great way to integrate a session management UX into the product directly.

martintrojer:
Thanks!

**Resuming PR sessions.** Turns out it isn't complicated: the persisted session already carries `pr_session_key` with the full `ForgeRepository` and number, so resume can hand that straight to `spawn_pr_open` — the same path the Pull Requests tab uses, spinner and all. 

I'll push PR resume here, then follow up with `dd` + confirm + the unhide toggle

agavra (approved):
Awesome change :) 🔥

PR #667, opened 2026-08-31
Fix Comment mode help alignment

nfvelten:
Closes #647

The Comment mode help section used hand-padded key labels, so descriptions for the longer key combinations started at different columns. Add a display-width-aware key column for all Comment mode rows, including the `comment_vim` row.

Validation:
- `cargo fmt --check`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo test`
- 1553 library tests, 1 binary test, and doc tests passed.

PR #673, opened 2026-09-02
Pan the help popup with h/l

### LINKED ISSUE #672: horiztonal scroll in help menu
This LGTM, but since the help popup doesn't have horizontal scroll capabilities yet (h/l) it makes some of the descriptions cut off on more narrower window with no way of seeing them. 

If you have time it'd be great to follow up this change with that!

_Originally posted by @agavra in https://github.com/agavra/tuicr/pull/667#pullrequestreview-5092317844_

nfvelten:
Follow-up to #667, closes #672.

The help popup renders unwrapped lines, so on a narrow terminal the descriptions get cut off at the right border with no way to reach them. This adds horizontal scrolling on `h`/`l` (and left/right arrows), matching the diff and file tree.

The render pass records the widest help line and the viewport width, so the offset clamps at the point where the longest description ends and does nothing when everything already fits. Title now reads `Help (j/k scroll, h/l pan, / search)`.

Tested by driving the TUI in a 60-column pty: before, `;h/;l  Move focus left/right (sid` is truncated; after four `l` presses the full `Move focus left/right (side-by-side...)` is visible. Unit tests cover the key mapping and the offset clamping.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1650,11 +1650,31 @@ impl Default for FileTreeFilter {
 #[derive(Debug, Default)]
 pub struct HelpState {
     pub scroll_offset: usize,
+    pub horizontal_offset: usize,
     pub viewport_height: usize,
-    pub total_lines: usize, // Set during render
+    pub viewport_width: usize,
+    pub total_lines: usize,    // Set during render
+    pub max_line_width: usize, // Set during render
     pub(crate) searchable_lines: Vec<String>,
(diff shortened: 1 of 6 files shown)


agavra (approved):
Super, thanks for the quick turnaround on this one 🙏

nfvelten:
No problem, glad to help. Thanks for the merge! I'll keep an eye on the issue list for more I can pick up.

PR #676, opened 2026-09-02
Add export.session_header to drop the session slug

### LINKED ISSUE #664: Feature request: optionally suppress the session slug when exporting to the clipboard
The `## Session` slug preamble seems to be the one thing we can't turn off using the `[export]` block in the tuicr configuration.

It seems to just confuse the agents [I use Codex] so I'd like to just turn it off.

Maybe something like:
  ```toml
  [export]
  session_header = false
  ```

nfvelten:
Closes #664.

`## Session: <slug>` was the only export line without a config key, so there was no way to trim it for agents that read it as noise. Adds `session_header` to the `[export]` block, defaulting to `true` — exports stay byte-identical until it is set.

```toml
[export]
session_header = false
```

Documented in `docs/CONFIG.md` next to the other export keys.

Verified end to end: with a draft comment in a local session and `--stdout`, the default export still opens with `## Session: nfvelten/tuicr@…`, and with `session_header = false` it opens directly on `I reviewed your code…`. Unit tests cover both the markdown gating and the config parse.


agavra (approved):
Thank you @nfvelten for the contribution! This looks good to me.

PR #677, opened 2026-09-03
feat(config): add editor key to override $EDITOR for e/:edit

yuhua99:
## Problem

Coming from tools like lazygit, I expected to configure the editor in `config.toml`.

Tuicr only reads `$EDITOR`, which means stuffing the environment variable through wrappers or aliases just to control which editor the `e` / `:edit` handoff opens.

## Proposal

Add a top-level `editor` config key:

```toml
editor = "nvim"
editor = "code -w"   # args allowed; -w keeps tuicr blocking on windowed editors
```

When set, it overrides $EDITOR.

## Implementation

- EditorCommand::from_env takes the override as a parameter and resolves editor config key → $EDITOR → vi in a small resolve_editor helper
- App carries editor_override from config into the editor handoff
- docs updated: docs/CONFIG.md, docs/KEYBINDINGS.md, README.md, help popup

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,7 @@ ignore_whitespace = false    # ignore all whitespace in local VCS diffs
 appearance = "system"        # or "dark" / "light"
 mouse = true
 leader = ";"                  # configurable prefix for leader shortcuts
+editor = "nvim"               # editor for `e` / `:edit`; overrides $EDITOR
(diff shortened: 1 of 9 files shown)


agavra (approved):
Thank you @yuhua99 for the contribution! This looks good to me.

PR #678, opened 2026-09-03
fix: avoid keyboard probe for piped stdin

### LINKED ISSUE #675: tuicr hangs at 100% CPU when reading diff from stdin
`When` piping a diff into `tuicr`, it hangs indefinitely and consumes ~100% of one CPU core.

For example:

```sh
git diff | tuicr
```

The same happens with Jujutsu:

```sh
jj diff --from xm | tuicr
```

Both `git diff` and `jj diff --from xm` work normally when executed without piping into `tuicr`.

### Expected behavior

`tuicr` should either:

* open and display the diff received from stdin, or
* exit with an error if reading diffs from stdin is not supported.

Instead, it enters what looks like a busy loop and keeps one CPU core at ~100%.

### Additional context

Even if stdin input is currently unsupported, the 100% CPU usage and indefinite hang seem like a bug.

### Reproduction

```sh
git diff | tuicr
```

or:

```sh
jj diff --from xm | tuicr
```

### Environment

```text
 ■ gear-picker HEAD * 22:34 ❯ tuicr --version
tuicr 0.24.0
 ■ gear-picker HEAD * 22:34 ❯ git --version
git version 2.55.0
```

atirna:
## Summary

Skip the keyboard enhancement probe when standard input is piped.

## Verification

- before: `git diff | tuicr` kept running and consumed a CPU core
- after: `git diff | tuicr` exits with a terminal-input error instead of spinning
- `cargo fmt --check`
- `cargo build`
- `cargo test`

Fixes #675

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::io::{self, Write};
+use std::io::{self, IsTerminal, Write};
 use std::sync::mpsc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thank you @atirna for the contribution! This looks good to me.

PR #680, opened 2026-09-03
docs(skill): guide agents through seeding comments correctly

martintrojer:
This targets the second workflow the skill already names — agent review of a patch — where the agent files its findings into tuicr with `review add` and the human then opens the TUI to read them. The review lands on the diff, anchored to the lines it is about, instead of arriving as a wall of chat text the human has to map back onto the code themselves. They page through it with `j`/`k`, mark hunks reviewed, reply, delete the findings they disagree with, and export what survives. The human is reviewing the review, in the tool they already use for reviewing. That flow only works if the agent anchors comments to the right lines, and today nothing in the skill tells it how.

Three failure modes on that path. Each succeeds silently, so nothing steers away from it.

**Reconstruct The Diff** (new section). An agent needs the same diff the user sees before it can pick line numbers, and the slug says which one. Adds the scope-per-slug-segment table and the `<base>~1` detail: range endpoints are inclusive, so diffing from `<base>` silently drops the first commit. Get this wrong and every line number is off, which is exactly how findings end up anchored to unrelated code.

**`--repo` caveat** (into Attach To A Session). `[]` with exit 0 also means "not a repo root" — a subdirectory returns it too. Read as "nothing open", an agent gives up while a session is running.

**Verify seeded comments** (into Add Agent Comments, next to the `--input` shape it depends on). A line past the end of the diff stores, prints back, and exits 0, but never renders — so the agent reports success on a review the human cannot see. Also: `review comments` reports no author, so the ids from `add` are the only way to tell agent comments from the user's later.

## Test plan

- Field shapes read off `src/review_cli.rs` and exercised against the built binary: the documented flat `--input` form round-trips, an array is a parse error, `path` is not an alias for `file`
- Range recipe measured here. For `commits/82e9508..b55e218`, `<base>~1..<head>` gives 15 files / 306 insertions; `<base>..<head>` gives 9 / 155 — short by exactly commit `82e9508` (6 files / 151)
- `[]` exit 0 reproduced for repo root, subdirectory, and nonexistent path
- Out-of-range line reproduced: line 9999 in a 3-line file returns the stored comment, exit 0

### DIFF
--- a/skills/tuicr/SKILL.md
+++ b/skills/tuicr/SKILL.md
@@ -49,11 +49,15 @@ If the user's intent is ambiguous, ask which workflow they want.
    local and PR sessions by owner/repo. Each row carries a `kind` (`local` or
    `pr`) and a usable `slug`. Use `--all` when you don't know the repo.
 
+   `[]` with exit 0 also means "not a repo root" — a subdirectory returns it
+   too. Pass the root, then `--all`, before concluding nothing is open.
+
 3. Choose the session:
    - If the CLI clearly reports exactly one relevant active session with
      `"active": true`, attach to it.
    - If multiple sessions are active, or the correct session is not clear, ask
-     the user which slug to use.
+     the user which slug to use. One repo can hold a worktree and a
+     commit-range session at once, and adding to the wrong one exits 0.
    - If the user provided a slug or session JSON path, use it directly.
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thank you @martintrojer for the contribution! This looks good to me.

PR #682, opened 2026-09-03
fix(cli): warn when review add --type is not a configured type

### LINKED ISSUE #681: review add accepts any --type, silently creating an unconfigured comment type
`tuicr review add --type BOGUS` succeeds and exits 0. The comment is stored with `comment_type: "BOGUS"` and exports as `**[BOGUS]**`.

`CommentType::from_id` ([src/model/comment.rs:92](https://github.com/agavra/tuicr/blob/main/src/model/comment.rs#L92)) maps any non-empty, non-`none` string to `Custom(id)`. That is right for the TUI, where ids come from validated `comment_types` config, but `review_cli.rs` never loads config, so the CLI has nothing to check an id against. A typo like `--type isue` is indistinguishable from a real custom type.

## Repro

```bash
tuicr review add --session <slug> --type BOGUS "test"
```

```json
{
  "comment_type": "BOGUS",
  "lifecycle_state": "local_draft",
  ...
}
```

Exit 0. On export, `export_comment_type_label` ([src/output/markdown.rs:535](https://github.com/agavra/tuicr/blob/main/src/output/markdown.rs#L535)) finds no matching definition and falls through to `comment_type.as_str()`, so the agent handoff document carries `**[BOGUS]**` — outside the ISSUE/SUGGESTION/NOTE/PRAISE vocabulary the receiving agent is told to expect.

## Why it matters for agents

The `tuicr` skill tells agents to pass `--type issue` and friends. A mistyped id is accepted, badged wrong in the TUI, and exported as an unknown type, with no signal at any step.

## Possible fix

Have the CLI load `comment_types` and reject an unknown `--type`, listing the configured ids. That would mean a config read on the `add` path, which it currently avoids — so an alternative is to keep accepting it and warn on stderr. Happy to send a PR for whichever you prefer.

Found while documenting CLI seeding for the agent skill (#680).

martintrojer:
Fixes #681.

`CommentType::from_id` maps any non-empty id to `Custom`. That is right for the TUI, where ids come from validated `comment_types` config, but the CLI had nothing to check against — so `--type isue` was stored, exported as `**[ISUE]**`, and badged as an unknown type, all at exit 0.

## Why warn instead of reject

Two reasons, and I would rather flag them than have them found in review.

`comment_types` is unset by default, and under that default every id but `none` is undefined. Rejecting unknown ids would make `--type issue` an error on a stock install — which is exactly what the agent skill tells agents to pass. Second, a scripted caller should not start failing because someone edited a config it does not control.

So: an absent or empty `comment_types` means no opinion and no warning. A configured list that omits the id gets a warning naming the valid ids. The comment is written either way and the exit code stays 0.

If you would rather have a hard error, or a `--strict-type` flag, say so and I will rework it — the placement is the same either way.

The warning goes to stderr, so stdout stays parseable JSON.

`add` already read config for the `username` fallback, so the check reuses that read rather than adding a second one; `resolve_cli_author` now takes the config instead of loading it.

## Test plan

- `cargo test`: 1585 passed, 0 failed. `cargo fmt --check` and `clippy --all-targets` clean (one pre-existing `bitbucket/bkt.rs` warning, present on main)
- Five new unit tests: warns and names the valid ids for an unconfigured id, silent for a configured one, silent when `comment_types` is absent or the config is default, silent for `none`, and the comment is still written when the type is unconfigured
- Manually, with `comment_types = [note, issue]`:
  - `--type issue` → nothing on stderr
  - `--type isue` → `Warning: comment type 'isue' is not configured; known types: note, issue` on stderr, stdout still valid JSON (verified with `json.load`), exit 0
  - No config file: both `--type issue` and the `none` default stay silent

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -205,6 +205,13 @@ and it is appended to the end of the Tab cycle when they are, so you can always 
 untyped. An untyped comment never renders a `[TYPE]` tag, a badge, or a legend entry (file-level
 comments still keep their `File-level:` marker on submit).
 
+### `tuicr review add --type`
+
+When `comment_types` is configured, `tuicr review add --type <id>` warns on stderr if `<id>` is not
+one of them, listing the ids that are. The comment is still stored and the command still exits `0`,
+so scripted callers keep working. With `comment_types` unset there is nothing to check against and
+no warning is emitted.
+
 ### Replacement semantics
 
 `comment_types` is a full replacement of the _configured_ types. If you define 2 types, those 2 —
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -214,6 +214,8 @@ pub enum ReviewCommand {
 
         /// Comment classification. Defaults to `none` (no type, no `[TYPE]`
         /// prefix); pass a type configured via `comment_types` to classify.
+        /// An id absent from a configured `comment_types` warns on stderr but
+        /// is still stored.
         #[arg(long = "type", value_name = "TYPE", default_value = "none", value_parser = non_empty_comment_type)]
         comment_type: String,
 
--- a/src/review_cli.rs
+++ b/src/review_cli.rs
@@ -92,7 +92,14 @@ fn add_comment(
     let request_parts = build_add_request_parts(options)?;
     let target = request_parts.target;
     let comment_type = CommentType::from_id(&request_parts.comment_type);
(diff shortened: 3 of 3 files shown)


agavra (approved):
Thank you @martintrojer for the contribution! This looks good to me.

PR #684, opened 2026-09-03
fix: stage deleted files with libgit2

### LINKED ISSUE #638: Staging removed file fails
Steps to reproduce:

1) Remove any previously tracked file in the repo
2) Open tuicr
3) Hit `r` to mark the file
4) Type `:stage`
5) `Git error: Could not find 'xxx' to stat: No such file or directory`

Expected behaviour: I should be able to stage the removed file.

mvanhorn:
Update the existing `stage_file` primitive to distinguish a path that still exists in the repository worktree from a tracked path that has been removed. Continue using the current index add operation for present files, but remove an absent tracked path from the index before performing the existing index write so the deletion becomes staged. When a reviewed, tracked file has been removed from the working tree, `:stage` reaches the libgit2 staging adapter and fails because that adapter unconditionally asks the index to add the missing path. The Git CLI backend already handles this case through `git add -- <path>`, so the defect is specific to the default libgit2 backend.

A tracked and committed file is deleted from the worktree, then staged through the libgit2 primitive: the call succeeds and the file is absent from the index relative to `HEAD`, representing a staged deletion; An existing new or modified file is staged: the current add-to-index behavior continues to succeed and the path remains present in the index.

Fixes #638

AI was used for assistance.

### DIFF
--- a/src/vcs/git/staging.rs
+++ b/src/vcs/git/staging.rs
@@ -5,7 +5,14 @@ use crate::error::Result;
 
 pub fn stage_file(repo: &Repository, path: &Path) -> Result<()> {
     let mut index = repo.index()?;
-    index.add_path(path)?;
+    if repo
+        .workdir()
+        .is_some_and(|workdir| workdir.join(path).exists())
+    {
+        index.add_path(path)?;
+    } else {
+        index.remove_path(path)?;
+    }
     index.write()?;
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thank you @mvanhorn for the contribution! This looks good to me.

PR #685, opened 2026-09-03
Print a review-exit summary line to stderr

joshgummersall:
## Summary

An agent driving `tuicr` from a pane can't currently tell "reviewed
everything, had nothing to flag" apart from "quit without looking" — both
show up as an empty `tuicr review comments` result. This adds a
`ReviewSession::comment_count()` helper (deduping the count logic that
previously lived only in `persistence/manifest.rs`) and prints one line to
stderr whenever the TUI exits:

```
tuicr-summary: reviewed <reviewed>/<total> files, <n> comment(s) added
```

`skills/tuicr/SKILL.md` is updated so an agent checks `reviewed_count` against
`file_count` before treating a zero-comment result as inconclusive: equal
counts means "reviewed, nothing to flag" and should be accepted; a shortfall
still triggers the "confirm with the user" fallback.

## Sample output

Real output from a `tuicr -r <revset>` run reviewing this PR's own 4 changed
files, exited with `ZZ`:

```
$ tuicr -r upstream/main..print-review-summary
tuicr-session: joshgummersall/tuicr@print-review-summary/commits/5e793fe..29345a6
tuicr-summary: reviewed 4/4 files, 0 comments added
```

## Test plan
- [x] `cargo test` — 1581 passed, 4 ignored, 0 failed
- [x] `cargo build --release`, manually ran the built binary against this
      branch's own diff and confirmed the `tuicr-summary:` line above

### DIFF
--- a/skills/tuicr/SKILL.md
+++ b/skills/tuicr/SKILL.md
@@ -155,10 +155,17 @@ seconds and compare comment IDs with the previous result. Read immediately when
 the user says comments are ready. Stop polling once the user says the review is
 done or your tooling would block other work.
 
-If the result is empty, ask whether the user saved comments in the intended
-session or whether another active session should be selected. If the review may
-have continued while you were working, rerun `tuicr review comments` before
-claiming completion.
+An empty result does not by itself mean the review didn't happen. On exit,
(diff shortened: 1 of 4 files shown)


agavra (approved):
Thank you @joshgummersall for the contribution! This looks good to me.

PR #688, opened 2026-09-04
Add q_quits config option

### LINKED ISSUE #686: Opt-in config to restore bare `q` as quit
0.25.0 (#630, `6fa287a`) turned bare `q` into a no-op that prints a "use `:q`" reminder. Agreed as the default - vim's normal-mode `q` records a macro. Asking for an opt-in key to get the old behaviour back:

```toml
q_quits = true   # default false = today's behaviour
```

Gate on the four `Action::QuitHint` arms in `src/input/keybindings.rs` (`map_normal_mode`, `map_visual_mode`, `map_commit_select_mode`, `map_submit_action_picker_mode`). `:q` / `ZZ` / `ZQ` unaffected. `ZQ` covers me today, but `docs/KEYBINDINGS.md` says the hint will be removed in a future release, after which there is no way back. Happy to PR it.

Separately: the 0.25.0 changelog says "**config:** Add confirm_quit to gate the bare q quit (#630)". No such key exists. #630 proposed one, the merge removed the binding for everyone instead, and git-cliff generated the entry from the stale commit title. That line is what sent me looking for a config option.

tuicr 0.25.0 (homebrew-core), macOS.

nfvelten:
Fixes #686

## Summary
- add the `q_quits` config option, defaulting to the current hint behavior
- make bare `q` quit only when enabled in normal, visual, commit-select, submit-action-picker, and file-tree modes
- document the option and keybinding behavior

## Testing
- `cargo fmt -- --check`
- `git diff --check`
- `cargo test --no-run`
- `cargo test q_quits`
- `cargo test should_map_q_to_quit_hint_instead_of_quitting`
- `cargo build`
- `cargo test --quiet -- --skip should_discover_worktree_with_relativeworktrees_extension`
- manual TUI smoke test with `cargo run --quiet -- --no-update-check`

The full suite passed except that the libgit2 worktree test `should_discover_worktree_with_relativeworktrees_extension` hangs locally, so it was skipped in the final complete run.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -33,6 +33,7 @@ show_reviewed = true
 mouse = true
 leader = ","
 comment_vim = false
+q_quits = false
 comment_tab_width = 4
 wrap = false
 relative_line_numbers = false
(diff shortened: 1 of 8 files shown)


agavra (approved):
Thank you @nfvelten for the contribution! This looks good to me.

PR #690, opened 2026-09-05
fix(cli): include comment author in review output

MattKotsenas:
# fix(cli): include comment author in review output

## Summary

This PR includes each persisted comment's `author` in the JSON emitted by
`tuicr review add` and `tuicr review comments`. Callers can distinguish comment
authors after reading a session instead of retaining the IDs returned when each
comment was added.

Comments added through the CLI use a nonblank `--username` or JSON
`username`/`author`, then a nonblank configured `username`, then `"user"`.
Comments written in the TUI use the configured `username`, or `"user"` when it
is unset. Agents should pass an explicit username when attribution must
distinguish their comments from the user's.

Open PR #680 currently says `review comments` reports no author. This PR adds
that missing field, so the statement in #680 will need updating.

### DIFF
--- a/docs/REVIEW_CLI.md
+++ b/docs/REVIEW_CLI.md
@@ -90,6 +90,8 @@ Target flags:
 - add `--line <n>` for a line comment
 - add `--end-line <n>` for a range comment
 - use `--side old|new` for inline comments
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thank you @MattKotsenas for the contribution! This looks good to me.

PR #691, opened 2026-09-05
fix(forge): return GitLab MR commits oldest-first

### LINKED ISSUE #660: Wrong diff is shown for commits in gitlab mr
Here is my sample repository which I use for this issue: https://gitlab.com/postolowicz/test. 
It contains one sample mr which we will use as example: https://gitlab.com/postolowicz/test/-/merge_requests/1

Here is git log view of this repo checked out locally:
```
64ad539 (origin/sample-mr) fixup!
e52022c Edit README.md
3fef430 (HEAD -> main, origin/main, origin/HEAD) Initial commit
```

When I run `tuicr mr 1` and select only first commit:

Image

"second change" is shown.

And when I select second commit both changes are present:

Image

For the reference, when I run `tuicr -r 3fef430..64ad539` everything is shown correctly:

Image

Image

Also notice: commits order in "Commits" section is different

Versions:
```
ubuntu@test:~/test$ tuicr --version
tuicr 0.24.0
ubuntu@test:~/test$ glab --version
glab 1.115.0 (c3612c8de)
```

atirna:
## Summary

GitLab MR commits now load oldest-first, matching what the inline commit selector expects.

GitLab's `merge_requests/:id/commits` endpoint returns newest-first while the other forge backends return oldest-first (the `ForgeBackend` contract — Bitbucket already reverses its newest-first rows in its backend). The GitLab backend passed the rows through, so every commit index was mirrored: selecting the first commit of the sample MR in #660 showed the second commit's diff instead. One `commits.reverse()` in the GitLab backend, same as the Bitbucket one.

This also fixes the commits order in the "Commits" section being reversed compared to `git log` — same root cause.

## Verification

- before, on current `main`: `cargo test forge::gitlab::glab::tests::should_return_commits_oldest_first` — failed, commits came back newest-first
- after: same command passes
- `cargo test` — all green
- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- also checked live that `gitlab.com/api/v4/.../merge_requests/1/commits` returns newest-first rows (the reporter's sample MR) and GitHub's `pulls/:n/commits` returns oldest-first, so the other backends are unaffected

Fixes #660

### DIFF
--- a/src/forge/gitlab/glab.rs
+++ b/src/forge/gitlab/glab.rs
@@ -429,6 +429,9 @@ where
                 break;
             }
         }
+        // GitLab lists MR commits newest-first; the trait contract is
+        // oldest-first.
+        commits.reverse();
         Ok(commits)
     }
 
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thank you @atirna for the contribution! This looks good to me.

PR #692, opened 2026-09-05
feat(gerrit): add Gerrit Code Review as a supported forge

### LINKED ISSUE #536: Feature Request: Support for Gerrit as a review target/source
At $work we use Gerrit for our code review system. I'm loving using tuicr for my local changes/claude integration, but it'd be awesome to read change requests from coworkers directly from Gerrit for review and submit comments/suggestions.

Gerrit provides an SSH-based CLI:

  https://gerrit-review.googlesource.com/Documentation/cmd-index.html

Although the `query, `review` commands are probably most important for tuicr:

  https://gerrit-review.googlesource.com/Documentation/cmd-query.html
  https://gerrit-review.googlesource.com/Documentation/cmd-review.html

anicke:
Closes #536.

Adds Gerrit Code Review as a supported forge. Changes open, browse, and submit like pull requests: a change number is the PR number, and `:submit` maps Comment / Approve / Request changes / Draft onto Gerrit reviews.

The issue suggested Gerrit's SSH CLI (`gerrit query` / `gerrit review`). This uses the REST API over HTTPS instead — it needs no `ssh` round-trip per call, returns structured JSON without the CLI's `--format json` quirks, and works with the same HTTP password users already have for `git push`. Gerrit ships no companion CLI the way GitHub, GitLab, Bitbucket, and Azure DevOps do, so `GerritBackend` is the first backend that talks to a forge directly rather than shelling out.

## What works

- Listing open changes, with `r` toggling to changes needing your review
- Opening a change by number, `https://host/c/<project>/+/<n>` URL, or legacy `#/c/<n>` URL
- Full diff rendering, file context expansion, and existing comments — including outdated ones from earlier patch sets
- Submitting reviews: message, `Code-Review` vote, and inline comments with multi-line ranges, in one POST
- Draft reviews, via Gerrit's per-comment drafts endpoint

## Notes on the implementation

**Detection.** Gerrit is always self-hosted with no reserved domain, so `parse_gerrit_remote_url` gates on the canonical SSH port `29418`, a `GERRIT_URL` naming the host, or a hostname containing "gerrit". It sits ahead of the GitHub catch-all but behind every host-gated parser, so GitHub Enterprise remotes on neutral hosts still fall through to GitHub.

**Project paths.** A Gerrit project is a path, not `owner/repo`, so `ForgeRepository::gerrit` packs the parent path into `owner` and the last segment into `name` — the same trick Azure uses for `org/project`. Single-segment projects fall back to the host as owner, since an empty owner produces an unparseable `ge:` slug.

**Review scope.** The `r` toggle queries `attention:self -owner:self` rather than `reviewer:self`, which would also match changes already voted on and handed back to the author. That needs Gerrit 3.3+.

**Diffs come from a local clone.** `git diff base..head` after fetching `refs/changes/NN/CCCC/P`. Gerrit's `/patch` endpoint is ordered by git's rename-aware diff queue while `/files/` is sorted by path, so the two cannot be zipped positionally. The fetch writes `FETCH_HEAD` only and moves no ref in the user's repo — enforced, not assumed: the ref is parsed into its change and patch-set numbers and rebuilt from those, so a server-supplied value cannot reach `git fetch` as an option or as a `<src>:<dst>` refspec.

**Auth.** A Gerrit HTTP password in `GERRIT_USERNAME` / `GERRIT_PASSWORD`; anonymous reads still work. Credentials are never sent over plaintext `http://` — that combination is refused outright, with no opt-in, since a Gerrit HTTP password also authorizes `git push`. A plain-HTTP `GERRIT_URL` is therefore anonymous-only.

## Not covered yet

- Patch-set comparison (reviewing PS2 against PS3)
- `Verified` and other custom labels — only `Code-Review` is voted on
- Submitting/merging a change from within tuicr
- Atomic draft submits. Gerrit's `Create Draft` endpoint takes one comment per request and has no batch equivalent, so saving a draft review is one `PUT` per comment. A failure part-way leaves the earlier drafts on the server while tuicr keeps every comment locally; nothing is lost, and the error names how many landed. Publishing a review is a single request and carries `omit_duplicate_comments`, so a retry after a lost response will not double-post inline comments.

## Testing

45 tests across `src/forge/gerrit/`, `src/slug.rs`, and `src/forge/mod.rs`, plus manual verification against a Docker-deployed Gerrit instance. `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (1626 passing) are all green.

See `docs/GERRIT.md` for setup.


src/forge/gerrit/api.rs
```diff
@@ -0,0 +1,1586 @@
...
+        // A bare `git fetch <remote> <ref>` writes FETCH_HEAD only — it creates
+        // no branch and moves no existing ref in the user's repo.
+        let _ = run_command_output(
+            "git",
+            Some(root),
+            [
+                "fetch",
+                "--quiet",
+                remote.as_str(),
+                pr.head_ref_name.as_str(),
```
agavra: Security: `head_ref_name` is populated from Gerrit's response and is passed to `git fetch` without validation. Git still parses option-looking values in this position—I verified that `--upload-pack=...` is treated as an option—and a value containing `:<destination>` can act as a refspec that updates a local ref, so the “FETCH_HEAD only” guarantee is not currently enforced. Could we validate that this is exactly the expected `refs/changes/NN/<change>/<patchset>` shape (including the current change number and a numeric patch set), reject `:` and leading `-`, and use an argument separator if supported? I think this should be fixed before merge.

src/forge/gerrit/api.rs
```diff
@@ -0,0 +1,1586 @@
...
+/// API never speaks SSH and a schemeless value is rejected outright (`http:
+/// invalid format`). Userinfo is dropped there, and so is a `29418` port —
+/// that is Gerrit's *SSH* port and means nothing over HTTPS. Any other port is
+/// kept: someone who writes `review.internal:8443` meant that port, and
+/// silently discarding it would be worse than the schemeless case this
+/// normalization exists to fix.
+fn configured_base_url() -> Option<String> {
+    let raw = non_empty_env(URL_ENV_VAR)?;
+    let trimmed = raw.trim_end_matches('/');
+    if trimmed.starts_with("http://") || trimmed.starts_with("https://") {
```
agavra, line 254: Security question: when both Gerrit credentials are set and `GERRIT_URL` explicitly uses `http://`, the Basic authorization header will be sent in plaintext. The docs currently advertise plain HTTP as supported. Could we either refuse authenticated HTTP, or require an explicit insecure opt-in and call the risk out prominently? Anonymous HTTP seems less concerning.

src/forge/gerrit/api.rs
```diff
@@ -0,0 +1,1586 @@
...
+    ) -> Result<()> {
+        let path = format!("/changes/{}/revisions/{}/drafts", pr.number, pr.head_sha);
+        let review_body = (!request.body.is_empty())
+            .then(|| json!({ "path": PATCHSET_LEVEL, "message": request.body }));
+        let inline = request.comments.iter().map(|comment| {
+            let mut payload = comment_input(comment);
+            payload.insert("path".to_string(), json!(comment_path_key(comment)));
+            Value::Object(payload)
+        });
+        for payload in review_body.into_iter().chain(inline) {
```
agavra: Functional edge: this sequence isn't atomic. If the nth `PUT` fails, earlier drafts have already been created remotely, but `create_review` returns an error and tuicr keeps all local comments as local drafts. Retrying can duplicate the successfully created subset. Is there a way to batch these, or at least report or track partial success so a retry is safe? If Gerrit has no batch endpoint, documenting the limitation may be enough for the first version.

agavra (commented):
Disclosure: These review comments were prepared with OpenAI GPT-5.6-Sol. I (Almog) don't have any prior experience with Gerrit and don't have a Gerrit server available for end-to-end testing, so this review is based on code inspection, the test suite, and local Git behavior. The overall direction makes sense; the patch-set ref validation comment is the one I consider blocking.

src/forge/gerrit/api.rs
anicke: Confirmed and fixed in d0d98ab.

You're right that the "FETCH_HEAD only" claim was asserted rather than enforced. Rather than
validating the string, I parse the ref into the two numbers Gerrit composes it from and format it
back out of those integers:

    let mut parts = pr.head_ref_name.strip_prefix("refs/changes/")?.split('/');
    let shard = parts.next()?;
    let change: u64 = parts.next()?.parse().ok()?;
    let patch_set: u64 = parts.next()?.parse().ok()?;

The change number has to be the change being fetched and the shard has to be the one Gerrit
derives from it (`change % 100`). What reaches git is digits and slashes by construction, so
there's no `:` and no leading `-` left to inspect for. A ref that doesn't match refuses to fetch
rather than being passed through.

On the argument separator: I tried it and left it out. `--end-of-options` only covers the option
half — I checked against real git, and `git fetch --end-of-options origin
refs/heads/main:refs/heads/pwned` still creates `refs/heads/pwned`. So it would cover one of the
two shapes you raised while adding a git 2.24 floor for that half, and the rebuild covers both.
The reasoning is in the doc comment so nobody adds it back thinking it was an oversight.

Test covers `--upload-pack=…`, `refs/changes/65/3965/2:refs/heads/main`, a leading `-`, a
mismatched change number, a wrong shard, the patch-set-less fallback ref, and empty.

src/forge/gerrit/api.rs
anicke: Agreed, fixed in 9f911e3 — and I went with the stricter of your two options.

Authenticated plaintext is now refused before the request leaves tuicr. Anonymous plaintext still
works, since there's no credential to leak. There's no opt-in escape hatch: a flag to send it
anyway is a flag to leak it, and a Gerrit HTTP password is worse than a typical API token because
it also authorizes `git push`.

Two things that made the hard refusal cheap. This only ever fires on a `GERRIT_URL` explicitly
written as `http://` — every other route to a base URL is normalized onto HTTPS — and the forge
hasn't shipped, so nobody is relying on the documented behavior yet.

I also considered exempting loopback so a local Docker Gerrit keeps working on a stock build, and
decided against it: "no plaintext credentials, ever" is a simpler rule to state and to document
than one with a carve-out, and a local instance can be reached over a tunnel or read anonymously.

`docs/GERRIT.md` now marks plain HTTP as read-only, explains why, and has the error in the
troubleshooting list. A test asserts the refusal message doesn't echo the password in either plain
or base64 form.

src/forge/gerrit/api.rs
anicke: Confirmed there's no batch endpoint — `Create Draft` takes a single CommentInput, and
`ReviewInput`'s `drafts` field only governs drafts already in the database. So the loop stays, and
I took your "document the limitation" fallback plus a better error, in 07f2dce.

I did start on the idempotent version (GET existing drafts, skip matching payloads) and dropped
it: nothing is actually lost on a partial failure — tuicr keeps every local comment — and the
duplicates are private drafts the author can delete in the web UI. Not worth a GET on every submit
plus content-matching that mishandles edit-then-resubmit. The error now names the split state
("Gerrit stored 2 of 3 draft comments before failing…") and it's recorded under MVP limitations.

Chasing this turned up something adjacent on the *publish* path, in 0df1f14: that POST is atomic
server-side, but its response isn't guaranteed to arrive, so a retry after a lost response
double-posted every inline comment. `omit_duplicate_comments` fixes it in one field. Separate
commit since you didn't ask for it — details in the commit message, happy to drop it.

agavra (approved):
Thank you anicke for the contribution! This makes sense to me.

PR #693, opened 2026-09-06
fix: scroll side-by-side diffs horizontally

### LINKED ISSUE #404: Horizontal diff scrolling doesn't work in split view
Hi!

I saw that the help mentions horizontal scrolling in diffs should be possible but it doesn't appear to be working. Is this a known issue?

Image

https://github.com/user-attachments/assets/e38c8394-99cd-40c6-92d9-5c15b497a5e7

In single diff mode it works when wrap is off

https://github.com/user-attachments/assets/f5b1f0a4-95af-4dab-8deb-a7be3723ae75

dkarter:
# Motivation

Side-by-side diff mode mapped `h` and `l` to horizontal scrolling, but each column truncated its content before calculating the offset. The view could not move when wrapping was off.

Closes #404 - i haven't been able to use side by side diff because of horizontal scrolling not working

# Summary of changes

- Calculate horizontal bounds per side-by-side content pane and pan both code columns while keeping gutters and the divider fixed.
- Keep the scroll position when opening line and range comments, with the editor fixed to the viewport.
- Add state and render-level regression tests for scrolling and comment entry.

# Testing

- Open a change with lines wider than each side-by-side column.
- Run `:diff`, then turn wrapping off with `:set wrap!`.
- Press `l` and `h`. Confirm that both code columns move while their gutters and center divider stay fixed.
- While scrolled right, open a line comment with `c`. Cancel it, select a range with visual mode, and open another comment. Confirm that both editors stay visible and the code remains scrolled.

https://github.com/user-attachments/assets/9a005593-85bf-4aee-b1d5-8938b4f44cb5

# Dependencies/Special Considerations

None.

### DIFF
--- a/src/app/comments.rs
+++ b/src/app/comments.rs
@@ -921,9 +921,9 @@ impl App {
 
     pub fn enter_comment_mode(&mut self, file_level: bool, line: Option<(u32, LineSide)>) {
         self.input_mode = InputMode::Comment;
-        // Snap horizontal scroll back to the left edge so the inline input
-        // box renders inside the viewport on long lines.
-        self.diff_state.scroll_x = 0;
+        if self.diff_view_mode != DiffViewMode::SideBySide {
+            self.diff_state.scroll_x = 0;
+        }
         self.comment_buffer.clear();
         self.comment_cursor = 0;
         self.comment_type = self.default_comment_type();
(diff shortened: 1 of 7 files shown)


agavra (approved):
Thank you @dkarter for the contribution! This looks good to me.

PR #701, opened 2026-09-08
Add review delete, clearc, and clear commands

### LINKED ISSUE #699: Add a delete/rm command to the review CLI
tuicr 0.25.0. The `tuicr review` CLI has `list`, `add`, and `comments` — but no way to delete a comment. The only deletion paths are TUI keybindings (`dd`, `:clearc`, `:clear`).

Problem: comments can become unreachable from the TUI. The local commit selector only shows the staged/unstaged rows when that side is non-empty (see #696), so once a side empties out, the session holding those comments cannot be reopened to delete them. The Sessions tab is meant to cover this, but resuming local staged/unstaged sessions is unreliable in practice.

Request: add a delete command, for example:

    tuicr review delete --session <slug> --comment-id <id>
    tuicr review clear --session <slug>   # all comments

or an `rm` subcommand, using the same `--session` resolution as the other commands (slug or JSON path) and deleting from the review/file/line comment buckets the same way `dd` does. This also gives agents and scripts the missing cleanup half of the `add`/`comments` loop.

jakubrpawlowski:
Add `tuicr review delete`, `clearc`, and `clear` (#699)

Three new review CLI commands close the add/comments loop for scripts and
agents: comments could previously only be deleted inside the TUI, and sessions
whose scope is currently empty were unreachable there at all.

- `delete --session <slug> --comment-id <id>` removes one comment from
  whichever bucket holds it (review, file, or line level)
- `clearc --session <slug>` clears all comments, keeping review marks
- `clear --session <slug>` clears comments and resets review marks
- all three remove the session itself when nothing is left (no comments and no
  review marks), matching TUI exit cleanup
- each writes JSON to stdout like the existing review commands

Supporting refactor, behavior-preserving: `CommentLocation` moved into the
model and the TUI `dd` removal logic extracted into
`ReviewSession::remove_comment`, now shared by `dd` and the CLI. The TUI path
is unchanged (full test suite green); the File arm additionally gained a
bounds check the inline code lacked.

6 new tests. `cargo fmt` and `cargo clippy` clean.

Closes #699

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,8 @@ tuicr --stdout | pbcopy
 ## Review session CLI
 
 `tuicr review` exposes saved sessions without opening the TUI. It can list
-sessions, add comments, and print stored comments for agent and script
+sessions, add, delete, and clear comments, and print stored comments for
+agent and script
 integrations. See [docs/REVIEW_CLI.md](docs/REVIEW_CLI.md).
 
 The TUI creates a persisted session file when a review target becomes active,
--- a/docs/REVIEW_CLI.md
+++ b/docs/REVIEW_CLI.md
@@ -91,6 +91,32 @@ Target flags:
 - add `--end-line <n>` for a range comment
 - use `--side old|new` for inline comments
 
+## Delete and clear comments
+
+Delete one comment by id:
+
+```bash
+tuicr review delete --session agavra/tuicr@main/worktree --comment-id <id>
+```
+
+The id is the `id` field reported by `tuicr review comments`. Deleting a
+comment that does not exist is an error.
+
+Clear every comment in a session:
+
+```bash
+tuicr review clearc --session agavra/tuicr@main/worktree
+```
+
+`clearc` keeps review marks. `clear` also resets them:
+
+```bash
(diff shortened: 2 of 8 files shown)


agavra (changes requested):
Thanks for the contribution @jakubrpawlowski - I think this makes sense. The Sol-high review found the following:

Problem: tuicr review delete can delete pushed or submitted comments because the lookup does not check their lifecycle.

Impact: The TUI intentionally treats those comments as read-only, so deleting their local record would make it diverge from GitHub.

Requested change: Restrict deletion to local_draft comments—or return a clear error for locked comments—and add a regression test.

jakubrpawlowski:
Fixed in 685fc55: deletion is restricted to local drafts. `remove_comment` refuses locked comments at the model level (`dd` unchanged), the CLI returns the TUI's "already pushed" message, and a regression test covers it. Suite green, fmt/clippy clean.

jakubrpawlowski:
@agavra friendly bump - this weekend I can address anything else you find

agavra (approved):
Thank you for updating it! Makes sense to me

PR #704, opened 2026-09-09
fix: persist reload reconciliation

MattKotsenas:
## Summary

With this PR, review markers invalidated by a reload stay invalidated after subsequent reloads or restarts.

When a full reload finds changed content, this PR persists the new content hashes and clears stale file and hunk review markers before installing the refreshed diff. It saves only this derived reconciliation, so unrelated comments and review choices remain dirty until the user explicitly saves them.

Review checkpoints:

1. Keep `persisted_session_snapshot` aligned with the session state loaded from or successfully written to disk. Existing-session read errors propagate instead of treating working state as the persisted baseline.
2. Centralize transitions that update the persisted snapshot, path, and file state.
3. Persist reconciliation performed by local reloads, including manual `:e`, editor return for reviews containing worktree changes, and post-staging reload.

Reloading a narrowed commit selection still updates only the current view. It does not change the saved session. Files that disappear from the diff keep their saved review state. If tuicr cannot save an update, it reports the error and leaves the current diff on screen.

The reconciliation persistence applies to normal local reloads, not background diff watching or PR reloads.

## Why

Previously, a reload could update hashes and invalidate reviewed state only in memory. The saved session could still contain the old hash and reviewed markers, allowing stale markers to reappear after a later reload or restart and make changed code look already reviewed.

## Testing

- Added regression coverage for persisted file and hunk invalidation.
- Covered dirty-state isolation, externally added comments, absent files, strict commit subsets, and save failures.
- Covered error reporting after a post-staging reload cannot persist reconciliation.


agavra (approved):
Thank you Matt for the contribution! This makes sense to me.

PR #705, opened 2026-09-09
docs: add a CLI reference and correct stale forge claims

joshvito:
I was reviewing the docs, to see what options were available via cli for dealing with reviews/sessions. Noticed some inconsistencies, so had a robit review and check docs matched the code. Here is that pass. -JV

Add docs/CLI.md covering the root command's flags, subcommands, environment variables, stderr markers, and exit codes. None of that had a reference page: --repo-url, --no-update-check, and TUICR_PROFILE were undocumented entirely, and --path, --file, and --version appeared only in generated changelog entries.

Also correct docs that contradicted the code. docs/KEYBINDINGS.md claimed to be a full reference while omitting 13 of the 57 commands in the registry, and its submit-support prose left out Gitea and Azure DevOps. docs/GITLAB.md and README.md said :submit draft was GitHub-only, but GitLab posts draft notes. docs/CONFIG.md's full example was missing four keys from its own options table.


agavra (approved):
Thanks @joshvito!

PR #706, opened 2026-09-10
docs(skill): correct review comment author guidance

MattKotsenas:
## Summary

This PR corrects the repository-managed tuicr skill to reflect that `tuicr review comments` includes each comment's `author`.

With this change, agents use `author` to distinguish their comments from the user's and retain `id` to identify an exact comment across reads. The skill's output-field list now includes `author` as well.

## Why

The skill incorrectly said that `review comments` did not report authors. That guidance contradicted the CLI output and `docs/REVIEW_CLI.md`, and forced agents to infer comment ownership from IDs alone.

### DIFF
--- a/skills/tuicr/SKILL.md
+++ b/skills/tuicr/SKILL.md
@@ -162,6 +162,7 @@ The command emits JSON. Each comment includes fields like:
 - `end_line`
 - `side`
 - `comment_type`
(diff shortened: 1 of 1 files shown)


agavra (approved):
Thank you Matt for the contribution! This looks good to me.

PR #712, opened 2026-09-14
perf: syntax highlight diff lines as they scroll into view

### LINKED ISSUE #613: Feature Request: Improve large diff rendering performance
## Problem

I tend to work and review large diffs when I refactor code (~4000+ lines and/or 50+ files).

`tuicr` renders the  syntax highlighting for the full diff in the main thread, which can cause the changeset to take ~5-10 seconds to render.

## Proposed Solution

I think we can speed up the rendering diff by:
1. Only syntax highlighting a sliding window of 3 screens at a time instead of the entire diff.
2. Backgrounding the syntax highlighting process to a background thread (or threads) so it doesn't delay the boot time of tuicr.

Once again let me know if you're interested @agavra. Happy to work on it in my spare time.

arjansingh:
Closes #613 

Large diffs can take several seconds to process before `tuicr` boots up.

A big reason for this is that `tuicr` syntax highlights all code, including code that is not visible.

This PR only highlights code as it enters a sliding window of the current screen plus one above and one below. The buffer helps minimize stuttering during scrolling. Highlighting is retained after code goes offscreen.

I did that primarily by defining and implementing a `LineColoring` type

We could further optimize memory usage by removing syntax highlights after the user is off screen, but that comes with a tradeoff of having to highlight it again.

## Time to Launch

This table measures time from launch to a drawn first screen, in a 200x50 terminal. This bench mark was runs 20 times per build. I'm showing the median value for each branch.

| Review Range | Files | Diff lines | `origin/main` | This branch | Faster by |
| --- | --- | --- | --- | --- | --- |
| v0.19.0..v0.23.0 | 127 | 38,096 | 1,274 ms | 168 ms | 7.6x |
| v0.22.0..v0.23.0 | 60 | 8,308 | 664 ms | 134 ms | 5.0x |
| v0.23.0..v0.24.0 | 24 | 1,854 | 158 ms | 112 ms | 1.4x |

## Memory and CPU

Ten runs per build for memory and CPU usage:

|                                             | Before   | After   |
| ------------------------------------------- | -------- | ------- |
| Opening the review, peak memory             | 100.1 MB | 62.0 MB |
| Opening the review, CPU                     | 1.25 s   | 0.15 s  |
| After reading the whole review, peak memory | 102.6 MB | 77.1 MB |
| After reading the whole review, CPU         | 5.26 s   | 4.96 s  |
| Worst single hunk spent coloring            | 179 ms   | 29 ms   |

## Benchmark Methodology

The benchmark uses tuicr history itself, specifically the commit range between `v0.19.0..v0.23.0`.

Instructions on how to run it are in the test file.


agavra:
Thanks @arjansingh - my philosophy is to keep rendering simple, the less state the better. I'm not sure who is reviewing a 38k line PR, and a ~2s load is totally reasonable in that world. 

The two other bugs I'd be happy to accept as individual PRs.

arjansingh:
Fair point, overcomplicating rendering is a legit concern. I remember our conversation from #613 and I'm not trying to introduce a lot of state tracking.

The 38k is the size of the benchmark diff I used. This PR is only about 1800 lines, 1500 net. Most of that is test cases.

Actual production code, including the bug fixes is about 600 lines. Most of that is just modifying plumbing. 

Actual new logic is about 200 lines.

I can split the fixes out. That would probably take this PR net down to ~1200 lines by my count.

Let me know, keep this together or split it up? Explain the changes here more or close this PR?

I've been using this for about 2 weeks locally, it's been a pretty impactful speed up to my workflow.

agavra:
Got it, let's start by splitting out the bugs into their own issues, then I can review this code in a bit more detail to understand the added complexity for just in time rendering.

arjansingh:
Sure, two bug PRs coming right up!

PR #726, opened 2026-09-17
perf(syntax): cap Oniguruma's backtracking budget

arjansingh:
Oniguruma did not handle Markdown files with inline code spans well.

The regex retry budget caused it to backtrack too much. Capping the budget at 100,000 steps eliminated this bottleneck and speeds up rendering time by about 85% for these cases.

Note: had to use `unsafe` to set this match limit.

Relevant constant being modified: https://github.com/kkos/oniguruma/blob/v6.9.10/src/oniguruma.h#L933


agavra (approved):
Thanks for tracking this down. The retry cap looks reasonable, and the full test suite passes with the latest main. Looks good to me.

PR #727, opened 2026-09-17
fix(diff): show the commit message for a commit picked from the list

arjansingh:
When a single commit was selected, I would see the commit message briefly flash then disappear.

With this fix the commit message should stay visible.


agavra (approved):
Thanks for fixing this. I added the reload cases too, so the message now survives :e and the diff watcher. The full test suite passes.

PR #729, opened 2026-09-18
feat: add automatic per-extension whitespace comparison

### LINKED ISSUE #700: smart/auto ignore whitespace
howdy! thanks so much for sharing this project! <3

as a follow-up to https://github.com/agavra/tuicr/pull/376 i wonder if there's scope here for automatically enabling/disabling sensitivity to whitespace based on the filetype or other criteria?

e.g. show whitespace changes in Python, YAML, etc where whitespace is significant

e.g. hide whitespace changes in JSON, JavaScript, Rust, etc where whitespace is insignificant

even if there wasn't any built-in filetype list, making it possible for the user to supply filename patterns or some other identifier would be amazing

cheers!

mvanhorn:
Extend configuration to accept the existing booleans plus the opt-in string `ignore_whitespace = "auto"`, keeping omitted/false as Normal and true as IgnoreAll, and add an `ignore_whitespace_overrides` TOML table mapping extensions to explicit booleans used only in auto mode. Supply a small documented built-in auto table (`json`, `js`, `jsx`, `mjs`, `cjs`, `ts`, `tsx`, `rs` ignore whitespace; unknown extensions, extensionless files, Python and YAML compare normally); user entries replace the decision for an extension, never invert their meaning, and describe this as a review heuristic rather than language-semantic equivalence. Issue #700 requests automatic whitespace comparison by file type, retaining whitespace changes for Python/YAML while allowing formatting noise to be ignored for JSON/JavaScript/Rust. Owner agavra explicitly welcomes a default configuration with user overrides; the reporter proposes extension-based configuration and warns that an exception list whose meaning reverses with another boolean would be confusing.

Config tests in `src/config/mod.rs`: omitted/true/false compatibility, auto defaults, explicit true and false overrides, empty table, inactive-table warning, invalid strings/types, malformed extension keys and normalized collisions. Existing invalid-boolean expectations must be updated deliberately for the expanded accepted syntax; Policy/selection tests in the new production module: uppercase extensions, multi-dot names, extensionless/dotfiles, non-UTF-8 paths where supported, destination-versus-source choice for renames/deletions, and reordered/missing ignored records. Verify original file order and that selected hunk coordinates/content hashes survive unchanged.

Fixes #700

AI was used for assistance.


agavra (approved):
Reviewed leniently; no remaining actionable findings. Integrated current main, added the missing pr_comments_visibility field to the test initializer, and moved the detailed whitespace configuration documentation from README.md to docs/CONFIG.md. Validation: cargo fmt, cargo clippy, and cargo test passed (1,785 tests passed, 4 ignored); documentation TOML examples also parse.

jokeyrhyme:
wow thanks!

mvanhorn:
Thanks for the merge, @agavra. Omitted and false still map to `Normal`, so this stays opt-in for anyone who does not want auto mode.

PR #730, opened 2026-09-19
feat(config): add pr_comments_visibility option

### LINKED ISSUE #658: Add config option to set default :comments visibility (outdated/resolved threads)
## Problem

Outdated and resolved GitHub review threads are hidden by default (PrCommentsVisibility::Unresolved). Running `:comments all` shows them, but visibility resets to unresolved for each new session - no way to persist a preferred default.

## Request

Add a config key, e.g. `pr_comments_visibility`, to set default value at startup:

```toml
   pr_comments_visibility = "all"  # "unresolved" (default) | "all" | "hide"
```

Would let users who always want to see outdated/resolved threads skip running `:comments all` every session.

I can submit pull request to resolve this issue

atirna:
## Summary

I added `pr_comments_visibility`, a config option that sets the default visibility of PR review comment threads: `"unresolved"` (the default, current behavior), `"all"` (resolved/outdated threads shown muted), or `"hide"`.

The configured value seeds each fresh PR session at the moment it is built. A session restored from persistence keeps the visibility saved with that review, whatever it is, and `:comments` still switches mid-session.

Fixes #658

## Why

Outdated and resolved threads are hidden by default, and the `:comments all` choice resets to unresolved for each new session, so anyone who always wants the full thread list has to run `:comments all` every time they open a PR.

## How

- `AppConfig` gains `pr_comments_visibility`, read with the existing `read_enum` (like `commit_order`), so an unknown value warns and is ignored.
- The resolved value rides `AppStartupOptions` onto `App::initial_comments_visibility`, and is applied where a fresh PR session is created (the PR-tab open, the new-head reopen, and the `tuicr pr <target>` startup path), before the persisted-session restore, which replaces the fresh session wholesale. That ordering is what keeps a saved visibility winning over the config.

## Verification

- `cargo test` (1753 passed, 0 failed), including new cases: a configured `all` seeds a fresh PR session through the real `open_pr_with_backend` seam, and a save/reopen cycle keeps a persisted visibility (saved `unresolved` stays `unresolved` under config `all`; saved `all` stays `all` under config `hide`); config parsing accepts the three values and warns on `"shown"` and on non-strings
- `cargo clippy --lib -- -D warnings`
- `cargo fmt --check`


agavra (approved):
Reviewed with a lenient bar; no actionable findings. The six focused configuration and session-visibility tests passed. Removed the detailed README configuration entry; the option remains documented in docs/CONFIG.md.

PR #732, opened 2026-09-20
feat: Add runtime theme picker with :theme command

wakuflair:
## Summary

Adds a runtime theme picker accessible via the new `:theme` command.

- `:theme` opens a picker modal listing available themes
- `j`/`k` live-previews a theme by reassigning `App::theme` per frame (cheap, no extra invalidation)
- `/` opens a filter draft (same UX pattern as the file-tree's `i`/`e`/`/` prompts), committed on Enter, discarded on Esc without touching the applied filter
- `Enter` on the picker keeps the preview and persists the choice to `config.toml` via a `toml_edit`-based edit that preserves comments and unrelated keys -- the only runtime command that persists to config, unlike other session-only `:set`/`:vim`/`:wrap` toggles
- `Esc` on the picker reverts to a `Theme` snapshot taken on entry and leaves the config file untouched
- `:theme <name>` applies and persists directly without opening the picker

Also updates the help popup, README, and `docs/KEYBINDINGS.md` to document the new command, and adds a new test suite (`src/app/tests/theme_picker_tests.rs`) covering picker navigation, filtering, and persistence.

https://github.com/user-attachments/assets/80f2adf7-f7ee-4785-9f0a-5b965de1e8ed


agavra:
Thanks for this @wakuflair, the live-preview picker is great!

I pushed one change before merging: `:theme` is now **session-only**. Picking a theme (from the picker or with `:theme <name>`) applies it for the current session, and the status line shows `set theme = "<name>" in config.toml to keep it`. The `config.toml` write and the `toml_edit` dependency are gone.

Why:
- **Consistency.** Every other runtime `:` command in tuicr (`:set`, `:wrap`, `:diff`, …) is session-only. `:theme` would have been the only command that writes the user's config.
- **Dotfiles.** Many people manage `config.toml` with stow, chezmoi or home-manager. Writing a temp file and renaming it over the config replaces a symlinked config with a plain file, which quietly cuts it off from their dotfiles repo.
- **Theme precedence.** Writing `theme` overrides `theme_dark`/`theme_light`, so one Enter in the picker would silently turn off automatic dark/light switching.
- **One less dependency.** `toml_edit` 0.25 pulled in a second copy of `toml_edit`/`toml_datetime`/`winnow`.

This matches how most other tools handle it:

| Tool | Runtime change | Writes to config? |
|---|---|---|
| **Helix** | `:theme <name>` and `:set` | No. It has `:config-open` to edit the file yourself and `:config-reload` to reload it. |
| **Neovim** | `:colorscheme` | No. You add it to init.lua yourself. |
| **Textual apps** (Posting, Harlequin, …) | The command palette changes the theme | No, unless the app adds its own saving. Harlequin takes `--theme` / config. |
| **Zellij** | In-app configuration screen, applied live | As I understand it, saving to the config is a separate, explicit step from applying for the session. |
| **btop** | Options menu | Yes, it saves automatically on exit (`save_config_on_exit`, on by default). This drew complaints from dotfiles users ([btop#1235](https://github.com/aristocratos/btop/issues/1235)). |

If people ask for a way to save the choice, we can add an explicit, opt-in save later (e.g. `:theme! <name>`) without changing the default.

agavra (approved):
LGTM, thanks for the contribution @wakuflair

PR #733, opened 2026-09-21
fix(comments): copy all remote comment types at cursor

### LINKED ISSUE #714: Feature Request: Copy comments from GitHub
Hey, i would like to copy comments from GitHub, tried Shift + y and it shows a message saying it copies only local commits.
Is there any reason this is not allowed ? If not i'm happy to help implement this. Thanks 
#713 ( sorry about opening another one, used my work account by mistake )

goiaciprian:
## Summary

  Fixes #714 

  Enable copying all displayed remote PR comments with Y, alongside existing local comments.

  Remote inline threads now retain which comment owns each rendered row, so copying a reply returns that reply rather than the thread root. Copying also supports:

  - Inline thread anchors and replies
  - Remote review summaries
  - Top-level PR conversation comments
  - Review-level remote threads without a diff-line anchor

  Anchor lookup now respects remote-comment visibility, preventing hidden resolved/outdated threads from being copied when they share an anchor with a visible thread.

  Remote thread row ownership is centralized to keep annotation, navigation, and rendering row counts aligned.

  ## Tests added

  - should_yank_rendered_remote_thread_when_hidden_thread_shares_its_anchor
      - Verifies an anchor copies the visible unresolved thread, not a hidden resolved one.

  - should_yank_reply_from_multiline_review_level_thread_footer
      - Verifies a multiline reply is copied from a review-level thread, including its closing/footer row.

  Existing/new copy tests also cover inline thread roots and replies, remote review summaries, and top-level PR comments.

  ## Validation

  - cargo fmt --check
  - cargo test
  - cargo test --lib


agavra (approved):
LGTM, thanks for the contribution @goiaciprian

PR #736, opened 2026-09-21
skill: put the tmux review pane where its caller is

### LINKED ISSUE #735: tmux wrapper opens the review in the focused pane, not the caller pane
When a coding agent (Claude Code in my case) runs `skills/tuicr/tuicr-wrapper.sh` from one tmux pane while I am working in a different pane or window, the review does not open next to the agent. It splits whatever pane I currently have focused and immediately takes focus, so my pane is blocked until I finish or discard the review.

Cause: `tmux split-window`, `tmux select-pane` and the `#{window_height}` lookup are all called without a target, so tmux applies them to the active pane of the client. `$TMUX_PANE` identifies the pane the wrapper actually runs in and should be used as the target.

Expected: the review opens beside the pane that launched it. Since the agent already notifies me when it finishes, I can switch to that pane when I have time. Focus should only jump to tuicr when I was already looking at the caller pane.

The zellij wrapper had the same problem and was fixed in #687 with `--near-current-pane`; the tmux wrapper was not updated.

Environment: tmux 3.6a, Linux, skill installed via `npx skills`.

nick4eva:
Fixes #735

`tuicr-wrapper.sh` called `split-window`, `select-pane` and `display-message` without a target, so tmux used the active pane of the client. When an agent launches a review from a pane the human is not looking at, the review lands in the pane of the human and steals focus.

Changes:
- Target `$TMUX_PANE` (the pane the wrapper runs in) for the split, falling back to the current pane if unset.
- Size the split from the caller `#{pane_height}` instead of `#{window_height}`.
- Only `select-pane` to the new pane when the caller pane is already active in its window; otherwise leave focus alone and log it.

This is the tmux counterpart of #687, which fixed the same behaviour for zellij.

Tested on tmux 3.6a: a split from a background pane opens in the window of that pane and the active window does not change.


agavra (approved):
LGTM, thanks for the contribution @nick4eva

PR #737, opened 2026-09-22
fix(navigation): target rendered hunk headers

### LINKED ISSUE #719: `[` / `]` jump to stale hunk positions after expanding hidden context in side-by-side view
Ran into this on 0.26.0 (installed via mise, Linux).

With side-by-side enabled (`:diff`), `[` and `]` move between hunks exactly like I'd
expect. But as soon as I expand hidden context, either with Enter for the 20 line chunk
or Shift-Enter for all of it, the jumps stop landing on hunks. The cursor ends up where
the hunk used to be before the context got expanded, so it's off by roughly however many
lines I just revealed. Collapsing the context again puts the jumps back on the real hunks.

Steps:

1. open a diff with a few hunks in the same file and collapsed context between them
2. `:diff` to switch to side-by-side
3. press `]` a couple of times, the cursor lands on the hunks correctly
4. expand the hidden context (Enter on a context row, or Shift-Enter for everything)
5. press `]` again

Expected: cursor moves to the start of the next hunk.
Actual: it lands short of it, at about the offset that hunk had before the expansion.
Same in the other direction with `[`.

Feels like the row positions used for the jump aren't recomputed after the expanded lines
are inserted. I only use side-by-side so that's where I noticed it, I haven't checked
whether unified behaves the same way.

hydraxman:
## Summary
- Derive hunk navigation targets from rendered `HunkHeader` annotations instead of counting raw diff lines.
- Keep `[` / `]` aligned after context expansion, paired changes and wrapped comments, while preserving filtering and single-file cross-file navigation.

## Problem
After expanding 20 context lines, the current implementation returns hunk positions `[1, 7]` although the displayed headers are at `[1, 30]`. This sends navigation into context rather than the next hunk. The annotation list already tracks the active layout and visibility rules; using it removes the separate, incomplete height calculation.

The same annotation-based approach appears in @pbjorklund's #475, which is now closed without merging. This PR is limited to #719 and its regression tests; it does not include the large-PR fallback from #475.

## Testing
On macOS, against base `fbc3cfce47e70e5ebba21b6df8de48daf16da029`:
- `cargo test --lib hunk_navigation -- --nocapture` — all four regressions fail with the base implementation and pass with this fix.
- `cargo test` — 1,808 passed, 4 ignored.
- `cargo build`, `cargo check`, `cargo clippy -- -D warnings`, `cargo fmt --all --check` — passed.
- Reviewed the complete two-commit range in the built tuicr TUI and exported the review.

Tests cover both layouts, single/multi-file views, expansion/collapse, wrapped comments, rendered header visibility after scrolling, and asymmetric cross-file jumps past a hidden file. Linux-specific CI remains for the upstream runner.

Fixes #719


agavra (approved):
LGTM, thanks for the contribution @hydraxman

PR #740, opened 2026-09-23
fix(editor): re-attach editor stdio to /dev/tty in --stdout mode

### LINKED ISSUE #739: `:edit` / `e` fails to attach a TTY when tuicr runs with `--stdout`
### Summary

When tuicr is launched with `--stdout`, opening a file via `:edit` (or the `e` keybinding) leaves a terminal editor with a non-TTY stdout, so `vim` / `nvim` refuse to render properly and print:

```
Vim: Warning: Output is not to a terminal
```

The editor is technically running, but it can't be used interactively.

### Why it happens

- `--stdout` mode redirects tuicr's own stdout to the file/pipe that captures the exported review. The TUI already draws on `/dev/tty` in this mode (see `src/main.rs`, the `let tty_output: Box<dyn Write> = if cli_args.output_to_stdout { … /dev/tty … }` block).
- `run_editor_from_tui` in `src/main.rs` calls `EditorCommand::run()`, which does `Command::new(...).status()`. `.status()` inherits stdio from the parent, so the spawned editor gets tuicr's file/pipe stdout, not the terminal.
- `vim` (and other terminal editors) detect the non-TTY stdout and drop into a batch-like mode with no real UI.

### Reproduction

The in-repo `skills/tuicr/` wrappers (the coding-agent integrations) rely on `--stdout` capture, so this reproduces trivially through them. Minimal manual repro:

1. Inside tmux:
   ```
   tuicr -w --stdout > /tmp/tuicr-out.md
   ```
2. Move the cursor onto any changed line and press `e` (or type `:edit`).
3. Observe `Vim: Warning: Output is not to a terminal` and a broken/black editor.

Reproduced with:
- tuicr `v0.26.0` at `fbc3cfc`
- vim 9.x and neovim (invoked via `alias vim=nvim`)
- Ubuntu 24.04, tmux 3.6a

### Expected

`e` / `:edit` should hand control to vim/nvim with a proper interactive terminal, just as it does when tuicr is run without `--stdout`.

### Proposed fix

When `output_to_stdout` is true, wire the spawned editor's stdin/stdout/stderr to `/dev/tty` (the same device the TUI already renders on). Concretely:

- Add `EditorCommand::run_on_tty()` alongside the existing `run()`, opening `/dev/tty` three times and passing them via `Stdio::from(...)` on Unix; fall back to plain `run()` on Windows.
- Expose a `run_editor_on_tty()` wrapper next to `run_editor()`.
- Pass `app.output_to_stdout` into `run_editor_from_tui` and pick the tty-attached variant when set. GUI/detached-editor path is unaffected (stdio is already `Stdio::null()`).

nick4eva:
When tuicr is launched with --stdout (e.g. by the tuicr-wrapper.sh used by coding-agent tmux integrations), its own stdout points at a file so the exported review can be captured. The TUI already draws on /dev/tty in that mode, but a terminal editor spawned via Command::status() inherits tuicr's stdio, so vim/nvim/hx see a non-TTY stdout and refuse to render properly ("Vim: Warning: Output is not to a terminal").

Add EditorCommand::run_on_tty (Unix; no-op fallback on Windows) that opens /dev/tty three times and wires stdin/stdout/stderr to it, plus a matching run_editor_on_tty wrapper. run_editor_from_tui takes the new output_to_stdout flag and picks the tty-attached path when set.

Closes #739


agavra (approved):
LGTM, thanks for the contribution @nick4eva

